PR #57, opened 2021-03-14
Add options to show Windows drive letters in uppercase, and to show colon after them.

louis-ad:
These options can be toggleed under Options -> Layout.

### DIFF
--- a/src/doublecmd.lpr
+++ b/src/doublecmd.lpr
@@ -158,7 +158,7 @@ begin
 
   DCDebug('Double Commander ' + dcVersion);
   DCDebug('Revision: ' + dcRevision);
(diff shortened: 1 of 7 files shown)


louis-ad:
Hello, I was looking for an Explorer replacement that has filtering of the listed files, and stumbled upon doublecmd. I didn't like that the drive letters were in lowercase, so I decided to look at it to see if I can change it. How interesting timing that this is the weekend you put the code on GitHub.

Comments are welcome, it's my first pull request.

j2969719:
not involved in development, and not sure how it works now after migration, but in the past, this project has а strict convention regarding commit messages: https://github.com/doublecmd/doublecmd/wiki/Guidelines-for-developers#Commiting_to_SVN

PR #59, opened 2021-03-15
ADD: Options to show Windows drive letters in uppercase and with a colon after the letter.

louis-ad:


src/platform/udrivewatcher.pas
```diff
@@ -533,7 +546,7 @@ class function TDriveWatcher.GetDrivesList: TDrivesList;
           with Drive^ do
           begin
             Path := DriveLetter + ':\';
-            DisplayName := DriveLetter;
+            DisplayName := Path;
```
alexx2000, line 549: Why Path? It should be "DriveLetter + OptionalColon" like below.

alexx2000:
Keep in mind that DC is a cross platform file manager. Usually I don't add platform specific options in GUI.
If such option added then it should be visible only on supported platform.

alexx2000:
I applied it  with small modifications and without GUI options.

PR #82, opened 2021-05-01
ADD: New supported "setup" parameter for internal command "cm_LoadFavoriteTabs".

denis-bisson:
Hi!
I took the time to make this video showing what the proposed modification would do:  https://youtu.be/cUBgEZdH16g
I will propose to add a little something in the help in a few minutes.

ADD: Modifiy the command "cm_LoadFavoriteTabs" so it would accept a parameter "setup=name" to intruct DC to load Favorite Tabs setup with name "name".
ADD: When the command "cm_LoadFavoriteTabs" has the parameter "setup=" with no value, it will invalidate the current Favorite Tabs selection so it won't save over it.
CHG: One new string in the language files.
ADD: Add the "actAddPlugin" declaration in the main form unit file that was missing for any reason.


alexx2000:
Hi, merged.

PR #92, opened 2021-05-28
UPD: Update build process

StefRe:
- Windows: use environment variable to set Lazarus path instead of modifying `build.bat` which is under source control (if merged I'll update the [relevant Wiki section](https://github.com/doublecmd/doublecmd/wiki/Development#from-command-line-windows) to first set the `LAZARUS_HOME` environment variable)
- add .gitignore file


StefRe:
@alexx2000 : I've seen that you have restricted editing of the Wiki to collaborators only, therefore I can't change it myself. So the second sentence [here](https://github.com/doublecmd/doublecmd/wiki/Development#from-command-line-windows) should be changed now to:  

> First you need the `lazbuild` utility of Lazarus to be somewhere in your PATH or you need to have an environment variable `LAZARUS_HOME` pointing to it (e.g. `set LAZARUS_HOME=d:\path\to\my\Lazarus`).

PR #131, opened 2021-07-03
ADD: More flexible quicksearch for chineses(Pinyin,Wubi,Shuangpin …)

fuwt:
fixd quicksearch for polyphonic character in Chinese.
use tcmatch.tal of total commander to get spellcode，so that tc users from china can quicksearch with custom code table, like Pinyin / Wubi / Shuangpin / Zhengma and more.


src/uPinyin.pas
```diff
@@ -0,0 +1,65 @@
...
+var
+  PINYINTABLE: pinyinarray;
+  PINYINTABLELOADED: boolean = False;
+
+procedure loadPinyinTable;
+var
+  f:file of pinyinarray;
+begin
+  if PINYINTABLELOADED then exit;
+  Assign(f, ExtractFileDir(Paramstr(0)) + '/tcmatch.tbl');
```
alexx2000, line 23: Use `ExtractFilePath(Paramstr(0)) + 'tcmatch.tbl'`

alexx2000:
Changing 'uMasks' unit will change mask checking everywhere which can be dangerous. I think 'TMask' should has option which enables Pinyin and it should be enabled in two places 'ufileviewworker.pas' and 'uorderedfileview.pas'.

fuwt:
actually,  new uMasks is superset of the old one. function 'PinyinMatch' is suit for English or other character, too.

fuwt:
I have added a option for function TMaskList.Matches to enable pinyinMatch. and disable it default. Now , it's maybe not dangerous.

src/uPinyin.pas
```diff
@@ -0,0 +1,65 @@
...
+var
+  PINYINTABLE: pinyinarray;
+  PINYINTABLELOADED: boolean = False;
+
+procedure loadPinyinTable;
+var
+  f:file of pinyinarray;
+begin
+  if PINYINTABLELOADED then exit;
+  Assign(f, ExtractFilePath(Paramstr(0)) + 'tcmatch.tbl');
```
alexx2000, line 23: No error checking. What is happen if file 'tcmatch.tbl' does not exists.

src/uPinyin.pas
```diff
@@ -0,0 +1,65 @@
...
+
+function PinyinMatch(a,b:UnicodeChar):boolean;
+var
+  i,code:word;
+  j:byte;
+begin
+  loadPinyinTable;
+  PinyinMatch := True;
+  if a = b then exit;
+  i := ord(a) - 19968;
```
alexx2000, line 40: I guess it should be in "if (Ord(c) >= 19968) and (Ord(c) < 40869) then" block.

Issue #130, opened 2021-07-02
Add Support for non-normalized file paths when assigning new directory path


TheNicker:
Currently changing directories using a non-normalized path such as: 'c:\windows/../windows` doesn't work, though it's completely legal and useful.

Sometimes when using source code, user  defined paths gets concatenated and this case can happen: `c:\windows/../windows` 

### Directly placing the path

![image](https://user-images.githubusercontent.com/4753283/124321083-c2302300-db85-11eb-833d-b8290d4fd3f8.png)

### Using the `cd` command

![image](https://user-images.githubusercontent.com/4753283/124321607-abd69700-db86-11eb-8e2e-bbad37a0aef6.png)

### Changing directory in the command line works great
![image](https://github.com/doublecmd/doublecmd/assets/4753283/a867a84b-9107-48c1-95c2-7e392f06cf4c)

PR #142, opened 2021-07-22
Added: support for non-normalized paths

TheNicker:
Solves issue: #130

The source of the issue is that Double commander doesn't normalize file paths properly, nor I found a good way to do it in the native Free pascal libraries.

This solution is similar to the C++ STL implementation and is windows specific by using the WINAPI call: `GetFullPathNameA` in this [line (681)](https://github.com/doublecmd/doublecmd/pull/142/files#diff-c04e2295a60f4dcb3eb4178838ff28a6422ace251f73885e53c8f0a3c19fe093R681)

### DIFF
--- a/src/fMsg.pas
+++ b/src/fMsg.pas
@@ -107,6 +107,7 @@ procedure TfrmMsg.FormKeyPress(Sender: TObject; var Key: Char);
 begin
   if (Key = #27) and (Escape >= 0) then
   begin
+    Key:= #0;
     iSelected:= Escape;
     Close;
   end;


TheNicker:
It would be best if someone have a code in pascal to contribute for normalizing directory paths. if not, a better cross platform solution will be something like the [STL of the GCC](https://github.com/gcc-mirror/gcc/blob/master/libstdc++-v3/src/c++17/fs_path.cc#L1672), but still should be translated to pascal which is not so straight forward. Might be worth adding it to the TODO list.

Normalization steps:
```
  - If the path is empty, stop.
  - Replace each slash character in the root-name with a preferred-separator.
  - Replace each directory-separator with a preferred-separator.
  - Remove each dot filename and any immediately following directory-separator.
  - As long as any appear, remove a non-dot-dot filename immediately followed
    by a directory-separator and a dot-dot filename, along with any immediately
    following directory-separator.
  - If there is a root-directory, remove all dot-dot filenames and any
    directory-separators immediately following them.
  - If the last filename is dot-dot, remove any trailing directory-separator.
  - If the path is empty, add a dot.
```

alexx2000:
It is enough to add "NewPath:= ExpandAbsolutePath(NewPath);" only. Function NormalizePathDelimiters + ExpandAbsolutePath done the work, so path "C:\Windows/../Windows" from example works fine.

TheNicker:
if you are meaning to this [line:](https://github.com/doublecmd/doublecmd/pull/142/files#diff-b3c2f0bea907204788a4b0a2da9e63faa391bdb309e84bcb5a7991c2bdd19dd6R119).

It solves more cases not all and only in the fileview header not in the command line as depicted in [this image](https://user-images.githubusercontent.com/4753283/124321607-abd69700-db86-11eb-8e2e-bbad37a0aef6.png) in issue #130.

For example it doesn't solve the case of `/\` or `\/` in the path name  (backslash followed by a slash and vice versa).
In general, this looks likes something that should be provided by the FPC RTL, similar to Delphi.

alexx2000:
Path with `/\` or `\/` is invalid. Why it should be processed? Windows Explorer does not support such path too.

TheNicker:
Indeed, windows Explorer doesn't support such a path too, Microsoft should fix it.

If you'll run `c:/\windows` in the command line it'll work. I even checked with another file commander such as [speed commander ](https://www.speedproject.com)and pasting ``c:/\windows`` into the directory header works there too.

The WINAPI function `GetFullPathName` defines that these kind of paths are valid.
In Linux shell Try `cd //\/\///\/\/\/etc` and it'll work, super legal.

I've encountered many times with these type of paths, usually when interop is involved with different libraries or when one of your developers is inconsistent with his preferred path separator, and he or she doesn't normalize paths as a habit.

avmaksimov:
What problem does it solve or what improvement does it make? May be better to close it?

PR #148, opened 2021-07-30
ADD: Feature [0002653] "Selected text only" enable if text highlighted

jack2m:
Option "Selected text only" in Editor enabled if text is highlighted


alexx2000:
Such feature will conflict with another one: autofill search dialog with selected text.

PR #160, opened 2021-08-22
ADD: Simple 32-bits checksum in available checksum functions

denis-bisson:
Used in retrocomputing to validate old ROM files.


alexx2000:
I think in this case standalone "dcchecksum16.pp" unit is not needed, function is very simple and should be implemented directly in the "TDCP_checksum16.Update(const Buffer; Size: longword)" method.

denis-bisson:
Hello Sir.
I've removed the "dcchecksum16.pp" file and made the checksum computation directly in the "Update" method  as you suggested. I also converted to 32-bits checksum instead of 16-bits, which was my first purpose anyway. Earlier, I've switched to 16-bits avoiding possible confusion for someone wanted to a CRC32 and would think "Checksum-32" would be the same... But on the other hand, users are not stupid, some will learn, we should not level down users, and anyway someone may confuse with CRC-16 anyway. So I've commited a second time. I hope I did it well, I've rarely made 2 commits in a row. I did not see how I could erase previous pull request to create a new one.

PR #169, opened 2021-08-26
Show current Dir in title bar, configurable in Miscellanious preferences

WhiteChairFromIkea:
2750u


alexx2000:
Static part of the title (GenerateTitle) should be called one time at start (as was before) but saved to variable and later used in "UpdateMainTitleBar".

src/fmain.pas
```diff
@@ -6171,6 +6159,40 @@ procedure TfrmMain.UpdateSelectedDrives;
...
+        dcVersion,
+        dcRevision,
+        dcBuildDate]
+    );
+  end;
+
+procedure TfrmMain.UpdateMainTitleBar;
+var sTmp: String;
+begin
+    if (fspDirectAccess in ActiveFrame.FileSource.Properties) and (gShowCurDirTitleBar = true) then
```
alexx2000, line 6181: I think it is better to use "if gShowCurDirTitleBar and (fspDirectAccess in ActiveFrame.FileSource.Properties) then"

src/frames/foptionsmisc.pas
```diff
@@ -37,6 +37,7 @@   TfrmOptionsMisc = class(TOptionsEditor)
     btnThumbCompactCache: TButton;
     chkDescCreateUnicode: TCheckBox;
     chkGoToRoot: TCheckBox;
+	chkShowCurDirTitleBar: TCheckBox;
```
alexx2000, line 40: Wrong indent

src/uglobs.pas
```diff
@@ -2495,6 +2497,7 @@ procedure LoadXmlConfig;
     if Assigned(Node) then
     begin
       gGoToRoot := GetValue(Node, 'GoToRoot', gGoToRoot);
+	  gShowCurDirTitleBar := GetValue(Node, 'ShowCurDirTitleBar', gShowCurDirTitleBar);
```
alexx2000, line 2500: Wrong indent

src/fmain.pas
WhiteChairFromIkea: fixed

WhiteChairFromIkea:
> Static part of the title (GenerateTitle) should be called one time at start (as was before) but saved to variable and later used in "UpdateMainTitleBar".

Fixed

src/fmain.pas
```diff
@@ -1039,22 +1042,6 @@ procedure TfrmMain.FormCreate(Sender: TObject);
     Result.OnDragOver:= @NotebookDragOver;
     Result.OnDragDrop:= @NotebookDragDrop;
   end;
-  function GenerateTitle():String;
```
alexx2000, line 1042: In this case no need to move function to another place.

WhiteChairFromIkea: In case `GenerateTitle()` declaration and implementations was placed in original position, I could not use it in `UpdateMainTitleBar`: "Error: (5000) Identifier not found "GenerateTitle"". Moving declaration + implementation near UpdateMainTitleBar solved that...

alexx2000: But you don't use it in "UpdateMainTitleBar", variable "sStaticTitleBarString" used here instead. "GenerateTitle()" is used only in "FormCreate" as is before.

WhiteChairFromIkea: Ok :)

WhiteChairFromIkea:
Force pushed last comments, sorry for a newbie problems, I hope this is the end :)

PR #254, opened 2021-11-07
MG : smaller thumbnail files as jpeg instead of png

mgholam:
Changed to jpeg thumbnails for all files.


Skif-off:
Maybe optional? If this is a real problem...

alexx2000:
We can't simply replace 'png' with 'jpg'. Users already can have thumbnails which saved as png. Also this patch are wrong because 'if' condition is useless in this case.  DC uses 'png' because it have better quality. I think we can do it optional and remove saved thumbnails when format changed.

mgholam:
- existing png files in the thumbnails folder work fine
- yes the if condition is useless (it was the minimal code change)
- technically png files are "lossless" but at the thumbnail resolution it makes no difference in quality

PR #294, opened 2021-12-14
Update string for 32bit and 64bit

bovirus:
@alexx2000 

Add strings (platform)  to show on top left for main installer window


alexx2000:
What installer will use it?

bovirus:
To show on top left of main window

"Double Commander x.x.x (32bit)" (for 32bit version)
"Double Commander x.x.x (64bit)" (for 64bit version)

Please check your email with my explanation about variables of Inno Setup in current version of Inno Setup scripts.

PR #377, opened 2022-03-20
fix: fix default Terminal.app location for macOS

calinrc:
The Terminal.app was moved from `/Applications` into  `/System/Applications` some time ago once Apple decide move internal apps in read-only fs.
Until a new build is created with this path change, macOS users that wants to launch terminal from Doublecmd top icon  can create a symbolic link from  `/System/Applications/Utilities/Terminal.app` to  `/Applications/Utilities/Terminal.app`

```bash
sudo ln -s /System/Applications/Utilities/Terminal.app /Applications/Utilities/Terminal.app
```

### DIFF
--- a/src/platform/uOSUtils.pas
+++ b/src/platform/uOSUtils.pas
@@ -20,7 +20,7 @@
 }
 
 unit uOSUtils;
- 
+
 {$mode delphi}
 
 {$IFDEF DARWIN}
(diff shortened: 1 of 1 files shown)


alexx2000:
DC supports old macOS versions too. So such change will broke it. The correct fix should check which path exists and use it, like [I done in the Linux](https://github.com/doublecmd/doublecmd/blob/master/src/platform/unix/udefaultterminal.pas).

PR #385, opened 2022-03-27
FIX: FPC 3.3 compatibility

larinsv:
The source file has been modified to allow doublecmd to be compiled with the current (trunk / main) version of FPC.  
At the same time, the project can also be compiled using FPC 3.2

### DIFF
--- a/plugins/wcx/unrar/src/unrarfunc.pas
+++ b/plugins/wcx/unrar/src/unrarfunc.pas
@@ -25,6 +25,9 @@
 unit UnRARFunc;
 
 {$mode objfpc}{$H+}
+{$ifndef VER3_2}
+{$modeswitch arraytodynarray}
+{$endif}
 {$include calling.inc}
 
 interface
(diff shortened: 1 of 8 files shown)


plugins/wcx/unrar/src/unrarfunc.pas
```diff
@@ -25,6 +25,7 @@
 unit UnRARFunc;
 
 {$mode objfpc}{$H+}
+{$modeswitch arraytodynarray}
```
larinsv, line 28: When compiling with FPC 3.2.2, a Warning occurs here:

> Warning: Illegal compiler switch "ARRAYTODYNARRAY+"

You can ignore this, or I can write something like this:

```pascal
{$ifndef VER3_2}
{$modeswitch arraytodynarray}
{$endif}
```

alexx2000: Yes, it is better to add IFDEF.

larinsv: Added IFDEF

vvch:
Unfortunately compatibility with fpc 3.0.4 was lost with this commit which is not good since this version of fpc is the only available one from the official repository in some still actual linux distributions, e. g. Ubuntu 20.04

PR #510, opened 2022-07-16
UPD: hold ALT key while dragdropping, popup ask menu

rich2014:
UPD: hold ALT key while dragdropping, popup ask menu (Copy/Move/Create Link/...), especially useful on Macs (due to lack of right mouse button)

### DIFF
--- a/src/platform/udragdropex.pas
+++ b/src/platform/udragdropex.pas
@@ -399,6 +399,8 @@ begin
     Result := DropMoveEffect
   else if ShiftState = [ssCtrl] then
     Result := DropMoveEffect
+  else if ShiftState = [ssAlt] then
+    Result := DropAskEffect
   else if ShiftState = [ssCtrl, ssShift] then
     Result := DropLinkEffect
   else


rich2014:
@alexx2000 

could you please check it?

alexx2000:
ALT key already used to force activation of external darg&drop. So I don't sure that it will not conflict. It is needed to check under all platforms.

rich2014:
1. it's just for drop destination, not the source, i guess there should be no conflicts
2.or should it be modified to be LCL_COCOA only?

alexx2000:
I will check how it work with other widgetsets.

ags1234:
Using doublecmd-1.1.0.r10306.i386-win32 on Windows 11.
Nothing different happens if I press Alt while dragging a file. It just calls for the Copy window, like not pressing Alt.
The only difference is the mouse tool tip. Without Alt is a big green plus sign, with Alt is a little grey plus.

rich2014:
could you please check with other widgetset? @alexx2000 

> I will check how it work with other widgetsets.

alexx2000:
Checked.

PR #511, opened 2022-07-16
FIX: get the app icon filename by native method

rich2014:
FIX: get the app icon filename by native method, support Binary format of Info.plist, replace the original HACK code


rich2014:
@alexx2000 
DARWIN only

PR #513, opened 2022-07-17
FIX: DARWIN: load the icon file content by native method

rich2014:
FIX: DARWIN: load the icon file content by native method, support all MacOS ICNS formats (bitmap/png/jpeg), and avoid limited image formats support of Lazarus.

tested on MacOS and Windows.

### DIFF
--- a/src/platform/unix/ushellcontextmenu.pas
+++ b/src/platform/unix/ushellcontextmenu.pas
@@ -42,6 +42,7 @@ type
     FFiles: TFiles;
     FDrive: TDrive;
     FUserWishForContextMenu: TUserWishForContextMenu;
+    FMenuImageList: TImageList;
     procedure PackHereSelect(Sender: TObject);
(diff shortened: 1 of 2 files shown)


rich2014:
@alexx2000 

image
image
image
image

src/platform/unix/ushellcontextmenu.pas
```diff
@@ -364,6 +365,8 @@ function TShellContextMenu.FillOpenWithSubMenu: Boolean;
       Result:= True;
       miOpenWith := TMenuItem.Create(Self);
       miOpenWith.Caption := rsMnuOpenWith;
+      menuImageList := TImageList.Create(nil);
+      miOpenWith.SubMenuImages := menuImageList;
```
alexx2000, line 369: ImageList is not freed anywhere.

rich2014: @alexx2000

Fixed, check it please.

PR #522, opened 2022-07-30
DEL: Obsolete and unnecessary keys in Info.plist

Kolcha:
`CFBundleSignature` is obsolete and should not be used
the rest keys are very specific, temporary, not documented, and not strictly required
everything works fine without them, so remove them


Kolcha:
closing due to out of interest

PR #527, opened 2022-08-03
ADD: HighDPI icons (for Qt5) in file panels

Kolcha:
first of all, this is **completely experimental changes**, I don't know Pascal (but know Qt pretty good, even its internals), just did it in similar way like other functions, so posting it as "draft" because strict review is required.

this was developed and tested on Debian Unstable running KDE with 200% scaling.

changes affect only Qt5 backend variant, anything else is not affected.

before the changes (take a look on the left side):
Screenshot_20220802_135808

after the changes (take a look on the left side, difference is much noticeable there):
Screenshot_20220802_161335

Lazarus sources from master branch are required to achieve the same effect on screenshots above. required change for Qt5 HighDPI support was committed pretty recently (~3 moths ago), so with older Lazarus these changes will not have any effect.

unfortunately, nothing can be done for other app icons (e.g. drives buttons) - these controls are provided by Lazarus and they don't use `QIcon` for icon painting, so they are doomed to be blurry... only Lazarus team can fix this.

file lists panels use custom painting on application side, so it can be "hacked" (what I did), fortunately Lazarus controls provide correctly configured surface (which can be recognized as HighDPI, thanks to recent commit), so icons can be painted correctly using `QIcon`.

### DIFF
--- a/src/platform/upixmapmanager.pas
+++ b/src/platform/upixmapmanager.pas
@@ -44,6 +44,9 @@ interface
 uses
   Classes, SysUtils, Graphics, syncobjs, uFileSorting, DCStringHashListUtf8,
   uFile, uIconTheme, uDrive, uDisplayFile, uGlobs, uDCReadPSD, uOSUtils
+  {$IF DEFINED(LCLQT5)}
+  , qt5, QtObjects
+  {$ENDIF}
   {$IF NOT DEFINED(DARWIN)}
   , uDCReadSVG
   {$ENDIF}
@@ -622,8 +625,11 @@ end;
 function TPixMapManager.CheckAddPixmap(AIconName: String; AIconSize : Integer): PtrInt;
 var
   fileIndex: PtrInt;
-  {$IFDEF LCLGTK2}
+  {$IF DEFINED(LCLGTK2)}
   pbPicture : PGdkPixbuf;
+  {$ELSEIF DEFINED(LCLQT5)}
+  qtIcon: QIconH;
+  iconName: WideString;
   {$ELSE}
   bmpBitmap: Graphics.TBitmap;
   {$ENDIF}
@@ -644,7 +650,7 @@ begin
         fileIndex := FPixmapsFileNames.Find(AIconName);
         if fileIndex < 0 then
           begin
-        {$IFDEF LCLGTK2}
+        {$IF DEFINED(LCLGTK2)}
             if not mbFileExists(AIconName) then
               begin
                 DCDebug(Format('Warning: pixmap [%s] not exists!', [AIconName]));
@@ -658,6 +664,16 @@ begin
               end
             else
               DCDebug(Format('Error: pixmap [%s] not loaded!', [AIconName]));
+        {$ELSEIF DEFINED(LCLQT5)}
+            iconName := UTF8ToUTF16(AIconName);
+            qtIcon := QIcon_Create(@iconName);
+            if not QIcon_isNull(qtIcon) then
+              begin
+                Result := FPixmapList.Add(qtIcon);
+                FPixmapsFileNames.Add(AIconName, Pointer(Result));
+              end
+            else
+              QIcon_Destroy(qtIcon);
         {$ELSE}
             {$IFDEF DARWIN}
             bmpBitmap := LoadImageFileBitmap(AIconName, AIconSize);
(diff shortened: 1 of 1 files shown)


ags1234:
Um... isn't the Qt version abandoned or not developed anymore? 
I don't see it here anymore: https://doublecmd.sourceforge.io/site/eng/snapshots.php

Kolcha:
Qt 5 version is available for Linux as it should https://sourceforge.net/p/doublecmd/wiki/Download/
more over, this the only option for Linux now, even it is not so good quality unfortunately... GTK2 is dead and almost gone from Linux world, and it is unusable on HighDPI screens due to its age. GTK3 is very unlikely to appear in Lazarus, but even that happen, it will be almost useless especially for such kind of apps - Double Commander uses screen space very rationally what completely contradicts modern trends (insanely huge controls with small and thin text inside) which GTK3 follows almost since it's begining. also GTK3 tends to disappear quite soon due to GTK4 which is considered as "good replacement" for GTK3.
Qt 5 also tends to disappear, but not so fast (thanks to the very buggy Qt 6 which was unusable before 6.3 even in base cases), but this is the best option that Lazarus provides.
unfortunately, to make this app great it should be rewritten in C++... but with such codebase it is almost impossible - too much code which is written in almost dead and very specific language which is pretty hard even to read compared to any widespread non JS-like languages. even more, even with deep understanding what happens in code, it can be noticed that it's design is so far to be even good, so the only option is write something completely from the scratch.
from other side, there are no so good (in cases of portability and features) GUI frameworks as Qt, but it is so far to be good option for such app. one "file view" widget with such convenient behavior as DC provides requires a lot of effort... I tried to use some ready to use classes provided by Qt and each time was unable to achieve the same behavior (and this excluding the possibility of customization), so very likely such widget should be written from scratch (what is completely not trivial task).
so, for now Qt 5 is what we have (on non-Windows platform). but it is also pretty unusable on macOS (due to Pascal - Qt integration for macOS in Lazarus).
--
Sent from Mail.ru app for Android Thursday, 04 August 2022, 05:37PM +02:00 from AGS  ***@***.*** :

JonnyTech:
> unfortunately, to make this app great it should be rewritten in C++...

@Kolcha  that is your personal opinion and one that many will strongly disagree with..

Kolcha:
@JonnyTech I agree, this is just my opinion, but let me explain why I think so. take into account only few requirements: cross-platform compatibility, access to system API (unfortunately this is required for this particular app), very good Linux support. let's look to possible common options:

- **Java** good cross-platform compatibility, has some UI toolkit, but very unlikely with good HighDPI support (at least based on what I saw). access to system API is pretty problematic (requires C/C++ library + JNI code). and it is very slow.... everyone knows how Java-based apps work, look for example Minecraft, Vuze, any JetBrains products, and famous AndroidStudio. and this is very problematic in packaging/distribution, especially for Linux.
- **any kind of JS / JS-like** just another monstrous app (150MB+) based on Chromium browser which many advanced users would like to avoid. also any UI toolkit "suffer" from modern trends with completely irrational screen space usage, this is unacceptable for this particular case. what with access to system API is a big question - such technologies are to high-level and usually don't assume doing something in platform-specific way. final app structure maybe acceptable for Win and macOS world, but not for Linux, look what inside Skype, Slack, Visual Studio Code or Atom packages for Linux (native packages, not snap/flatpack, such things even must not exist in Linux world, this is just a result of try to commercialize Linux somehow).
- **Go, Rust and so on "fancy" languages** no good UI toolkit, anything existing is just some kind of very poor quality "bindings" to C/C++ frameworks (aka Qt) or "adapters" for JS-based frameworks using TCP connections between backend and UI (completely unacceptable!). in terms of distribution, it requires a lot of exotic runtime, installing which because only one app doesn't worth it.
- **C#/.Net** maybe good option for Windows, but definitely not for cross-platform (even .Net runtime exists for many platforms, but it is very bad quality comparing to Windows). good UI toolkit (at least was, don't know what is there right now, don't use Windows, so didn't modern see .Net-based apps), but it at least had problems with HighDPI. access to system API requires some efforts, but yet pretty easy (at least on Windows, hard to what is on other platforms).
- **C++** good cross-platform compatibility (C++ compiler exists anywhere), few mature pretty good quality UI kits to choose from (Qt and GTK for example, Qt preferable), easy access to any system API because it is C (or at least "C-compatible"), easy way to distribute app, no specific runtime is required - all required runtime already exists in any system.

JonnyTech:
@Kolcha this is an issue tracker, not a place to discuss opinions of programming languages.

(BTW, both Double Commander and Total Commander are written in Pascal which means that it does have its merits)

Kolcha:
anyone interested may try this on Debian by installing my nightly build from there:
https://software.opensuse.org//download.html?project=home%3Anikoneko%3Adoublecmd&package=doublecmd-qt

Kolcha:
closing due to out of interest

Issue #341, opened 2022-01-31
Unable to toggle off reserve space for good


israelio:
On the options windows at the copy / move there is an option to toggle off the Reserve space - which slows down heavily the copy operation. It works fine but i am missing to turn it off for good via configuration or once and for all... toggle it off once and let it be remembered and not have to press the option each time and turn it off.

![image](https://user-images.githubusercontent.com/2522999/151867199-ecea48f9-0425-49d9-9059-145021056683.png)

Using Mac build v1.1.0 alpha build 10045 2022/01/06

alexx2000:
Really this dialog has "save" button (which saves current options as default):

![Untitled](https://user-images.githubusercontent.com/13401398/151914709-e1ee4b41-f93e-4473-a954-dd89756f9b25.png)

but it invisible due #340

israelio:
For the record - Mac Users ... open the option tab, click shift+tab which will bring you to the unreachable Save there option as default then Space and later enter and click on the start.... it will save the options as default 
This is a TEMP solution and from my opinion it should be fixed... as it is fixable

j2969719:
>as it is fixable

how exactly?

PR #543, opened 2022-08-11
FIX: COCOA: controls position error in CopyMoveDlg

### LINKED ISSUE #340: Some options are unreachable on the copy / move (F5/F6)
On Mac using v1.1.0 alpha build 10045 2022/01/06

When one click on the options button you cant get to the options in the bottom, the window is not resizable nor can scroll up/down the options there.

image

image

rich2014:
fix #340 and #341 
tested on MacOs and Windows

### DIFF
--- a/src/fcopymovedlg.lfm
+++ b/src/fcopymovedlg.lfm
@@ -263,7 +263,6 @@ object frmCopyDlg: TfrmCopyDlg
     Color = clForm
     ParentColor = False
     TabOrder = 3
-    Visible = False
     OnResize = pnlOptionsResize
     object grpOptions: TGroupBox
       AnchorSideLeft.Control = pnlOptions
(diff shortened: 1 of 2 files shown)


rich2014:
image

src/fcopymovedlg.lfm
```diff
@@ -10,6 +10,7 @@ object frmCopyDlg: TfrmCopyDlg
   ActiveControl = edtDst
   AutoSize = True
   BorderIcons = [biSystemMenu, biMaximize]
+  BorderStyle = bsSingle
```
alexx2000, line 13: Is it a mandatory change? Can it be done only for Cocoa (via code)?

rich2014: 1. it's not a mandatory requirement, not even for cocoa
2. it's just a convention, the dialog is fixed size like other dialogs in DC
3. CopyMoveDlg was originally set to be resizable, is there any special consideration?

alexx2000: Many dialogs in DC are resizable. Only some dialogs has fixed size. User may want to resize CopyMoveDlg if full target path is very long.

rich2014: THX.
form BorderStyle setting restored.
check it please.

PR #545, opened 2022-08-12
FIX: COCOA: "Select file name without extension when renaming" doesn'…

rich2014:
…t work


rich2014:
fix #296 "Select file name without extension when renaming" doesn't work

PR #550, opened 2022-08-17
FIX: COCOA: F2-Rename EditButton: VK_RETURN handler in IME state

rich2014:
1. when using F2-Rename in MacOS, RETURN KEY is incorrectly handled in IME input state. the intermediate typing text is returned as the result, then returns directly and incorrectly changes the filename.
2. the correct way is to receive IME text, and still in the Rename EditButton, just like other apps on MacOS and DC on Windows.
3. tested on MacOS and Windows.


rich2014:
image

rich2014:
image

rich2014:
image

rich2014:
@alexx2000 hi, Alex, check this PR please.

PR #553, opened 2022-08-19
FIX: F7-mkdir: VK_ESCAPE VK_RETURN in IME state

rich2014:
1. In Lazarus 2.2, it is no longer necessary to handle the Enter key in TForm.OnKeyPress
2. After removing the OnKeyPress code, the IME state can be handled correctly

### DIFF
--- a/src/fmkdir.lfm
+++ b/src/fmkdir.lfm
@@ -13,7 +13,6 @@ object frmMkDir: TfrmMkDir
   Constraints.MinHeight = 50
   Constraints.MinWidth = 350
   KeyPreview = True
-  OnKeyPress = FormKeyPress
   Position = poOwnerFormCenter
   LCLVersion = '2.2.0.4'
   object lblMakeDir: TLabel
--- a/src/fmkdir.pas
+++ b/src/fmkdir.pas
@@ -20,7 +20,6 @@ type
     procedure cbExtendedChange(Sender: TObject);
     procedure cbMkDirChange(Sender: TObject);
     procedure cbMkDirKeyUp(Sender: TObject; var Key: Word; Shift: TShiftState);
-    procedure FormKeyPress(Sender: TObject; var Key: Char);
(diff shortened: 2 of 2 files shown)


Skif-off:
> In Lazarus 2.2, it is no longer necessary to handle the Enter key in TForm.OnKeyPress

Does this mean that building packages for some Linux distributions will be dropped?
For example, Ubuntu/Ubuntu-based 18.04 (support until April 2023) and 20.04 (support until April 2025).

alexx2000:
This need to check with GTK2/Qt5. Removed code is a workaround for Lazarus bug.
Maybe it works with Lazarus 2.0.12 too. Anyway I plan to update Lazarus to 2.2.2.

rich2014:
1. frmMkDir is a regular dialog with Default Button (to handle VK_RETURN) and Cancel Button (to handle VK_ESCAPE), Lazarus' default KEY handling mechanism can handle it well
2. I think only in the special case such as #550 (F2-Rename with only one TEditButton), we need to handle VK_RETURN and VK_ESCAPE by ourself

rich2014:
could you please test with GTK2/Qt5? @alexx2000

rich2014:
> This need to check with GTK2/Qt5. Removed code is a workaround for Lazarus bug.
> Maybe it works with Lazarus 2.0.12 too. Anyway I plan to update Lazarus to 2.2.2.

**also the snapshot I downloaded from official website, shows that Lazarus 2.2.2.0 is already in use?**

Double Commander
Version: 1.1.0 alpha
Revision: 10317
Commit: 2dffc7af8
Build date: 2022/08/17
Lazarus: 2.2.2.0
FPC: 3.2.2
Platform: x86_64-Darwin-cocoa
OS version: Mac OS X 10.16.0

alexx2000:
I talk about Linux builds. It use Lazarus 2.0.12 under some Linux distributions.

PR #555, opened 2022-08-21
UPD: F7-mkdir: improve the layout (label example overlay buttons occasionally)

rich2014:
UPD: F7-mkdir: improve the layout, add space between controls.
and avoid label example overlay buttons when "Extended syntax" checked.

### DIFF
--- a/src/fmkdir.lfm
+++ b/src/fmkdir.lfm
@@ -1,6 +1,6 @@
 object frmMkDir: TfrmMkDir
   Left = 366
-  Height = 104
+  Height = 125
   Top = 429
   Width = 350
   ActiveControl = cbMkDir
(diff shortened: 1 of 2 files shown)


rich2014:
image

rich2014:
image

rich2014:
@alexx2000 could you please check it?

alexx2000:
In this case window change height on checkbox on/off. So far I don't know how to do it better.

rich2014:
how about this version? by aligning the OK button to the bottom of the form, and setting the form's
MinHeight, keep the height of the form unchanged

alexx2000:
No, it works incorrect. Controls can be much bigger and window autosize does not work in this case.
So I think it's better to use first variant but clear `lblExample.Caption` (in `RefreshExample`) by space symbol instead empty string.

rich2014:
set space instead of empty string, it works fine.
check it please @alexx2000 

> ... But clear `lblExample.Caption` (in `RefreshExample`) by space symbol instead empty string or ...

alexx2000:
Also need to add `RefreshExample` here:
```
    RefreshExample;
    cbMkDir.SelectAll;
    Result := (ShowModal = mrOK);
```

PR #570, opened 2022-09-03
FIX: COCOA: kaspathedit in IME state

rich2014:
1. when using PathEdit in MacOS, RETURN KEY and ESCAPE KEY are incorrectly handled in IME input state.
2. the intermediate typing text is returned as the result when RETURN KEY pressed, then returns directly and incorrectly changes the path. the correct way is to receive IME text, and still in the PathEdit, just like other apps on MacOS and DC on Windows.
3. PathEdit will be completely canceled when ESCAPE KEY pressed. the correct way is to ignore IME intermediate typing text only, and still in the PathEdit, just like other apps on MacOS and DC on Windows.
4. tested on MacOS and Windows.

### DIFF
--- a/components/KASToolBar/kaspathedit.pas
+++ b/components/KASToolBar/kaspathedit.pas
@@ -43,6 +43,13 @@ type
     FStringList: TStringList;
     FObjectTypes: TObjectTypes;
     FFileSortType: TFileSortType;
+{$IF DEFINED(LCLCOCOA)}
+    originalText: String;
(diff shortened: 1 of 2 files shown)


rich2014:
image

rich2014:
image

rich2014:
image

rich2014:
@alexx2000 hi, Alex, could you please check this PR?

PR #571, opened 2022-09-03
UPD: PathLabel: add highlight calling in MouseEnter

rich2014:
1. in PathLabel, the current version only highlights the path selected in the MouseOver Event.
2. when the mouse first enters the PathLabel, it may only trigger the MouseEnter Event, resulting in missing highlighting.
3. fix the problem by adding highlight related codes in MouseEnter Event.
4. tested on MacOS and Windows.

### DIFF
--- a/src/upathlabel.pas
+++ b/src/upathlabel.pas
@@ -249,6 +249,8 @@ begin
   if FAllowHighlight then
   begin
     Cursor := crDefault;
+    FMousePos := ScreenToClient(Mouse.CursorPos).X;
+    Highlight;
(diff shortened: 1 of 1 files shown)


rich2014:
https://user-images.githubusercontent.com/102933068/188275857-2d41ce71-13a8-4798-8780-3f327c71cb50.mov

rich2014:
https://user-images.githubusercontent.com/102933068/188275913-1c57ff35-c3fd-4a30-aaf2-b4e539e873dc.mov

PR #577, opened 2022-09-12
ADD: DARWIN: integrated MacOS Native QuickLook Preview (Ctrl-Q)

rich2014:
ADD: DARWIN: integrated MacOS Native QuickLook Preview (Ctrl-Q)
1. integrated native QuickLook Preview on MacOS, more formats supported and faster
2. only used in `Quick View Panel (Ctrl-Q)`, original implementation of DC used in `View (F3)`
3. tested on MacOS and Windows
4. at the same time, it supports MacOS to use the native QuickLook implementation, and other OS continue to use DC original implementation. In order to keep the code clear, this PR is divided into three commits.
5. please adopt `Rebase and merge` to keep commits individually

### DIFF
--- a/src/platform/unix/darwin/uquicklook.pas
+++ b/src/platform/unix/darwin/uquicklook.pas
@@ -28,13 +28,42 @@ unit uQuickLook;
 interface
 
 uses
-  Classes, SysUtils;
+  Classes, SysUtils, ExtCtrls,
+  MacOSAll, CocoaAll, CocoaInt,  QuickLookUI;
+
+type
+  TQLPItem = objcclass(NSObject, QLPreviewItemProtocol)
+    url: NSURL;
+    function previewItemURL: NSURL;
+  end;
+
+type
+  TDarwinQLControl = class
+  private
+    superpanel: TPanel;
+    superview: NSView;
+    view: QLPreviewView;
+  private
+    procedure setFilepath( filepath: String );
+    procedure OnResize( Sender: TObject );
+    function isOpened(): Boolean;
+  public
+    constructor Create( panel: TPanel );
+    destructor Destroy; override;
+    procedure open;
+    procedure close;
+    procedure resize;
+  public
+    property filepath: String write setFilepath;
+    property opened: Boolean read isOpened;
+end;
+
 
 implementation
 
 uses
-  DynLibs, FileUtil, Types, Graphics, MacOSAll, CocoaAll, uThumbnails, uDebug,
-  uClassesEx, uGraphics;
+  DynLibs, FileUtil, Types, Graphics,
+  uThumbnails, uDebug, uClassesEx, uGraphics, uMyDarwin;
 
 const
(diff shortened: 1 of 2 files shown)


rich2014:
pdf

rich2014:
excel

rich2014:
vector

rich2014:
movie

rich2014:
audio

Skif-off:
In the current implementation, this feature will disable any plugins in quick view mode. Maybe it's better to make it optional?

Enable by default, but without GUI option (because this feature is not cross platform), i.e. enable/disable only by direct editing `doublecmd.xml` (like [these parameters](https://doublecmd.github.io/doc/en/configxml.html)).

rich2014:
as a MacOS user, my thought is that there is little need to add the option.
there are obvious advantages to using the native quick look preview.
and the plugins still work in View (F3).
of course, it would not be difficult to add option to the config file if needed later.

> In the current implementation, this feature will disable any plugins in quick view mode. Maybe it's better to make it optional?
> 
> Enable by default, but without GUI option (because this feature is not cross platform), i.e. enable/disable only by direct editing `doublecmd.xml` (like [these parameters](https://doublecmd.github.io/doc/en/configxml.html)).

Skif-off:
> and the plugins still work in View (F3).

I believe you understand the difference between "View" and "Quick View" and what advantage the latter has :)

This feature looks very interesting and I'm glad you took the time, but I don't think it's good to radically disable part of the available functionality. In any case, it's your right :)

P.S. I am not a macOS user.

rich2014:
@alexx2000 hi,Alex, could you please check this PR about QuickView (QuickLook)

alexx2000:
It's a serious change. I will check it.

alexx2000:
I've been thinking a lot and I suppose that this implementation violates a principle that platform specific code should be maximally isolated (if possible) from cross-platform code, otherwise it will be unsupported hell. Also it disables cross-platform functionality. DC already has a several abstractions for such isolation. Some time ago under Windows I added similar function "Explorer Preview" as a [wlx-plugin](https://github.com/doublecmd/doublecmd/tree/master/plugins/wlx/preview/src). I guess that QuickLook can be implemented as a wlx-plugin too (which activated in quick view mode only). Similar plugin [existed](https://github.com/doublecmd/plugins/tree/master/wlx/quicklook) in the past but it was created when DC does not have a native Cocoa version (DC used Qt4) so it has a problems. But now with native Cocoa version such plugin should work without problems. Also it can be activated in normal view when needed (using lcp_forceshow). So I propose try to implement it as a wlx-plugin and then decide what to do.

rich2014:
thanks for your detailed reply.
it seems obvious that it is more suitable to implement as wlx-plugin.
i will learn wlx-plugin and reimplement it this weekend.

rich2014:
reimplemented in #603 as a wlx-plugin

PR #585, opened 2022-09-14
UPD: Refactor TEditButtonEx and TFileViewWithMainCtrl

rich2014:
1. refactor TEditButtonEx and TFileViewWithMainCtrl, the code related to IME input state
2. the code is simpler, more reasonable, and more readable (refer to KASPathEdit)
3. tested on MacOS and Windows

### DIFF
--- a/src/fileviews/ufileviewwithmainctrl.pas
+++ b/src/fileviews/ufileviewwithmainctrl.pas
@@ -109,6 +109,9 @@ type
     procedure edtRenameButtonClick(Sender: TObject);
     procedure edtRenameMouseDown(Sender: TObject; Button: TMouseButton;Shift: TShiftState; X, Y: Integer);
 
+{$IFDEF LCLWIN32}
+    procedure edtRenameKeyDown(Sender: TObject; var Key: word; Shift: TShiftState);
+{$ENDIF}
     procedure edtRenameOnKeyESCAPE(Sender: TObject);
     procedure edtRenameOnKeyRETURN(Sender: TObject);
   protected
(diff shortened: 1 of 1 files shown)


src/fileviews/ufileviewwithmainctrl.pas
```diff
@@ -1463,116 +1560,10 @@ procedure TFileViewWithMainCtrl.edtRenameExit(Sender: TObject);
...
-      end;
-
-{$IFDEF LCLGTK2}
-    // Workaround for GTK2 - up and down arrows moving through controls.
-    VK_UP,
-    VK_DOWN:
-      Key := 0;
-{$ENDIF}
-
-{$IFDEF LCLWIN32}
```
alexx2000, line 1555: This code should go to EditKeyDown.

rich2014: 1. In the committed version, the code related to LCLGTK2  has been moved into EditKeyDown (line 297-300)
2. the code related to LCLWIN32 should no longer be needed in Lazarus2.2, i have tested it. of course I'm not 100% sure about it yet

```
{$IFDEF LCLGTK2}
     // Workaround for GTK2 - up and down arrows moving through controls.
     VK_UP,
     VK_DOWN:
       Key := 0;
 {$ENDIF}

{$IFDEF LCLWIN32}
```

alexx2000:
OK, I will check.

alexx2000:
I found that this behavior is not depend on Lazarus version but on Windows version. It works correct with Windows 10, but does not work with Windows 7. So this code should stay but it can be changed like this:
```
{$IFDEF LCLWIN32}
    // Workaround for Win32 - right arrow must clear selection at first move.
    VK_RIGHT:
      begin
        if (Win32MajorVersion < 10) and (Shift = []) and (edtRename.SelLength > 0) then
        begin
          Key := edtRename.CaretPos.X;
          edtRename.SelLength := 0;
          edtRename.CaretPos := Classes.Point(Key, 0);
          Key := 0;
        end;
        FRenFile.UserManualEdit:=True; // user begin manual edit - no need cycle Name,Ext,FullName selection
      end;
     VK_LEFT:
        FRenFile.UserManualEdit:=True; // user begin manual edit - no need cycle Name,Ext,FullName selection

{$ENDIF} 
```

rich2014:
OK. check it please.

PR #600, opened 2022-09-23
UPD: the last part of the text in the TEdit is automatically selected, when selecting from AutoComplete ListBox in KASPathEdit

rich2014:
the last part of the text in the TEdit is automatically selected, when selecting from AutoComplete ListBox in KASPathEdit, more like on Windows.
tested on MacOS 12 and Window 11.

### DIFF
--- a/components/KASToolBar/kaspathedit.pas
+++ b/components/KASToolBar/kaspathedit.pas
@@ -30,7 +30,7 @@ interface
 
 uses
   Classes, SysUtils, LResources, Forms, Controls, Graphics, Dialogs, StdCtrls,
-  ShellCtrls, LCLType, LCLVersion, Strings
+  ShellCtrls, LCLType, LCLVersion
 {$IF DEFINED(LCLCOCOA)}
   , CocoaAll, CocoaWindows
 {$ENDIF}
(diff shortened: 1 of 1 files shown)


rich2014:
image

components/KASToolBar/kaspathedit.pas
```diff
@@ -367,6 +368,26 @@ procedure TKASPathEdit.CreateWnd;
...
+  Text:= newText;
+end;
+
+{$ELSE}
+
+procedure TKASPathEdit.setTextAndSelect( newText:String );
+var
+  start: Integer;
+begin
+  if Strings.strpos(pchar(newText),pchar(Text))<>nil then
```
alexx2000, line 382: You can use `Pos` function here, like:
`if Pos(Text, newText) > 0 then`

components/KASToolBar/kaspathedit.pas
```diff
@@ -367,6 +368,26 @@ procedure TKASPathEdit.CreateWnd;
   FAutoComplete:= not SHAutoCompleteX(Handle, FObjectTypes);
 end;
 
+procedure TKASPathEdit.setTextAndSelect( newText:String );
```
alexx2000, line 371: This variant is not needed, second variant should be without `IFDEF`. It will be used if native Windows autocomplete failed.

rich2014:
@alexx2000
thx. check it please.

rich2014:
@alexx2000 
could you please check this PR?

PR #608, opened 2022-09-26
UPD: DARWIN: enable MacPreview wlx-plugin

rich2014:
1. build MacPreview.wlx  with plugins/build.sh
2. install MacPreview.wlx with install/create_packages.mac


alexx2000:
Also need to add it into `src/platform/udefaultplugins.pas`, similar as Windows wlx-plugins.

PR #610, opened 2022-09-27
FIX: fix the problem of displaying wrong hint text in QuickViewPanel

rich2014:
fix the problem of displaying wrong hint text in QuickViewPanel
1. disable hint text in QuickView
2. tested on MacOS 12 and Windows 11


rich2014:
image

PR #616, opened 2022-10-01
delete *.lrs, *.res

rseleven:


alexx2000:
Only
`components/dcpcrypt/DCPciphers.lrs`
`components/dcpcrypt/DCPhashes.lrs`
`plugins/wcx/base64/src/base64wcx.res`
can be deleted. Other files are used and not auto-generated.

PR #620, opened 2022-10-02
FIX: COCOA: garbled when editing in FileViewHeader

rich2014:
FIX: garbled when editing in FileViewHeader:
1. this problem only occurs on MacOS and dark mode
2. in dark mode, the TEdit with border becomes transparent, and the PathEdit fails to cover the PathLabel
3. set PathEdit to opaque by BorderStyle:=bsNone
4. tested on MacOS 12 and Windows 11


rich2014:
image

rich2014:
image

Issue #207, opened 2021-10-19
Show filename in status bar when nothing is selected. [Feature Request]


corvus1:
In Full view currently when nothing is selected it shows "Selected 0 of..." which is not very useful. It'd would be nice to show current filename under the cursor like in FlatView or BriefView, because filenames are sometimes long and don't fit into the column.

corvus1:
Ping. Any thoughts on the proposition?

alexx2000:
It's useful, it shows file/folder count. At my system for example most of file names are fit to panel width.

corvus1:
Yeah, but when they don't, the only way to see the full name is mouse cursor hint, when you hover over it.

corvus1:
I'm not sure it's a great argument, but that's what Total Commander does.

alexx2000:
Total Commander works same way as DC by default.

corvus1:
Is that not the default?

![Screenshot_20220919_001504](https://user-images.githubusercontent.com/4068104/190926592-c1e07c5c-bfed-44ba-8aba-5cc0efa76213.png)

I didn't know that.
But anyway, I'd appreciate the option.

frithrah:
I would find it helpful too. I like to do everything with the keyboard so having to hover over a file with the mouse just to see the full name is a shame.

alexx2000:
> Is that not the default?

No. It depends on the `LongInStatus` key in wincmd.ini, by default `LongInStatus=0` .

corvus1:
I must've set it 20 years ago and forgot about it.

corvus1:
Just tested. Seems to work as expected. Thanks!

corvus1:
Maybe a cosmetic change to add [brackets] for directories? :-)
To look more like Total Commander.

PR #677, opened 2022-11-22
A cosmetic change for 09f989d4. Show [brackets] around directories names in the status bar if nothing is selected.

corvus1:
A bit of obsessive mimicking of Total Commander. The discussion is in the end of #207 
Please consider.

### DIFF
--- a/src/fileviews/ucolumnsfileview.pas
+++ b/src/fileviews/ucolumnsfileview.pas
@@ -738,29 +738,25 @@ end;
 procedure TColumnsFileView.UpdateFooterDetails(AInfo: Boolean);
 var
   AFile: TFile;
-  F : Longint;
 begin
   if gColumnsLongInStatus and (FSelectedCount = 0) and (not FlatView) then
   begin
     AFile:= CloneActiveFile;
     if Assigned(AFile) then
-    try
-    if AFile.IsNameValid then 
-      begin
-        F:=FileGetAttr(AFile.Name);
-        If (F<>-1) and ((F and faDirectory)<>0) then 
-          lblInfo.Caption := '['+AFile.Name+']'
+      try
+        if AFile.IsNameValid then begin
+          if AFile.IsDirectory then 
+            lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix
+          else 
+            lblInfo.Caption := AFile.Name;
+        end                       
         else 
-          lblInfo.Caption := AFile.Name;
-      end
-    else 
-      begin
-        if not AInfo then 
-          inherited UpdateInfoPanel;
+         if not AInfo then begin
+            inherited UpdateInfoPanel;
+         end;
+      finally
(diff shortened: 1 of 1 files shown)


alexx2000:
Wrong code, it should be same as [here](https://github.com/doublecmd/doublecmd/blob/master/src/ufilefunctions.pas#L217).

corvus1:
Oh. I didn't realize that. :-) Well, I'm unfamiliar with the codebase, but I didn't introduce any new dependencies, not a single `uses`.

I'll try to rework it, but I'll have to take a good long look into it. :-)

Skif-off:
See `<FolderPrefix>` and `<FolderPostfix>` [here](https://doublecmd.github.io/doc/en/configxml.html).

Simething like
```pas
      if AFile.IsNameValid then
      begin
        if AFile.IsDirectory then
          lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix
        else
          lblInfo.Caption := AFile.Name
      else if not AInfo then begin
        inherited UpdateInfoPanel;
      end;
```
without `F`?

corvus1:
@Skif-off 

Yes, I figured it out, quicker than I originally thought. The prefix and postfix come from the settings, and AFile has IsDirectory, so there's not reason to check it separately.

There's, however, one more problem I'm not sure how to solve, maybe you'll help me a little. The way I did it, it doesn't add brackets on DC start. When you start the program, the directory your cursor was on the last time shows in the status bar with no brackets, until you move the cursor, then it updates and shows the brackets.

So apparently that function doesn't get called on DC start, which might or might not indicate a bigger problem, actually. But to locate that I'd really have to familiarize myself with the way the whole thing works.

corvus1:
I pushed the correct implementation, and the other problem I was talking about is somehow... gone (?), fixed itself (?). IDK, I'll take it. Problem fixing themselves is nice. :-)

src/fileviews/ucolumnsfileview.pas
```diff
@@ -743,15 +743,20 @@ procedure TColumnsFileView.UpdateFooterDetails(AInfo: Boolean);
...
     AFile:= CloneActiveFile;
     if Assigned(AFile) then
-    try
-      if AFile.IsNameValid then
-        lblInfo.Caption := AFile.Name
-      else if not AInfo then begin
-        inherited UpdateInfoPanel;
+      try
+        if AFile.IsNameValid then begin
+          if AFile.IsDirectory then
```
alexx2000, line 748: It should be `if gDirBrackets and (AFile.IsDirectory or AFile.IsLinkToDirectory) then` as I mentioned above. It should has same display name as name on the file panel.

corvus1: Yeah, sorry for missing that. Fixed.

corvus1:
Thank you, everyone! My first contribution. :1st_place_medal:

PR #681, opened 2022-11-25
FIX: TFileViewNotebook.DestroyAllPages: algorithm of tabs.clear()

rich2014:
FIX: TFileViewNotebook.DestroyAllPages: algorithm of clear tabs

when switching Favorites, a large number of Tabs that will be removed are activated unnecessarily.
this is because in `TFileViewNotebook.DestroyAllPages()` call `Tabs.Clear()`. the Clear() algorithm will cause the next Tab to be activated time by time after removing the currently active Tab. even FCanChangePageIndex:=False doesn't work.

In some cases, it will cause LCL control exception. for example, in `Thumbnails View` mode, in the Destroy process, because the Tab is reactivated, MakeFileSourceFileList() is restarted, and then TFunctionThread is restarted to update data in the background. at this time, the LCL control has been destroyed, thus exception thrown.

it exists at least in Cocoa, because NSTabView cannot be in the state of missing active tab, otherwise an exception will be thrown.

it was fixed by fine-tuning the algorithm in DestroyAllPages(). remove the Tab in the same reverse order, but skip the active tab, and remove the active tab at the end, to avoid activating other tabs in DestroyAllPages().

after the patch, the exception has been fixed. and the speed of switching Favorites has been visibly improved because it avoids activating the Tabs to be removed. 

tested on MacOS 12 and Windows 11.

### DIFF
--- a/src/ufileviewnotebook.pas
+++ b/src/ufileviewnotebook.pas
@@ -540,7 +540,7 @@ var
   i: Integer;
 begin
   for i:=PageCount-1 downto 0 do
-    if i<>ActivePageIndex then RemovePage( i );
+    if i<>ActivePageIndex then inherited RemovePage( i );
   RemovePage( 0 );
 end;
 
--- a/src/ufileviewnotebook.pas
+++ b/src/ufileviewnotebook.pas
@@ -107,7 +107,6 @@ type
 
   TFileViewNotebook = class(TPageControl)
   private
-    FCanChangePageIndex: Boolean;
     FNotebookSide: TFilePanelSelect;
     FStartDrag: Boolean;
     FDraggedPageIndex: Integer;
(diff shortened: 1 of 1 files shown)


rich2014:
image

JonnyTech:
Is this an issue with the `.Clear` function that should be reported to the Lazarus team?

alexx2000:
In this case `FCanChangePageIndex` variable and `CanChangePageIndex` function override should be removed. They was added specially for `DestroyAllPages`.

rich2014:
> Is this an issue with the `.Clear` function that should be reported to the Lazarus team?

yes. and I have submitted several PRs to Lazarus before, some of them are approved, some are not yet, the review time is relatively long.

rich2014:
> In this case `FCanChangePageIndex` variable and `CanChangePageIndex` function override should be removed. They was added specially for `DestroyAllPages` and didn't need anymore.

removed.

alexx2000:
`DestroyAllPages` should call `inherited RemovePage` because overriden version makes sense only when not all tabs removing.

rich2014:
> `DestroyAllPages` should call `inherited RemovePage` because overriden version makes sense only when not all tabs removing.

done

alexx2000:
I think last call `RemovePage( 0 )` should be replaced too.

rich2014:
fixed

PR #682, opened 2022-11-26
Show link target in status bar.

corvus1:
Useful at times. Without it, the only way I found to quickly see where symlink is pointing is to go open the file properties.


corvus1:
Oh, geez, gihub made a `git mess`

I shouldn't have reused my fork. Is this going to be a problem?
If it is, I can delete the fork, make a new one, apply the changes, this time, hopefully, in one neat commit, and make a new pull request.

PR #683, opened 2022-11-26
FIX: QuickView: Not exiting PluginMode when receiving empty string in TfrmViewer.LoadFile()

rich2014:
Not exiting PluginMode when receiving empty string in TfrmViewer.LoadFile().
will cause multiple Plugins content to be overlaped.

```pascal
procedure TfrmViewer.LoadFile(const aFileName: String);
......
  if dwFileAttributes = faInvalidAttributes then
  begin
    ActivatePanel(pnlFolder);
    ExitPluginMode;  // missed before
    ----------------
    memFolder.Font.Color:= clRed;
    memFolder.Lines.Text:= rsMsgErrNoFiles;
    Exit;
  end;


procedure TQuickViewPanel.FileViewChangeActiveFile(Sender: TFileView; const aFile: TFile);
......
  except
    on E: EAbort do
    begin
      FViewer.Hide;
      FFirstFile:= True;
      Caption:= E.Message;
      FViewer.LoadFile(EmptyStr);
      ---------------------------
    end;
  end;
```

### DIFF
--- a/src/fviewer.pas
+++ b/src/fviewer.pas
@@ -696,6 +696,7 @@ begin
   if dwFileAttributes = faInvalidAttributes then
   begin
     ActivatePanel(pnlFolder);
+    ExitPluginMode;
     memFolder.Font.Color:= clRed;
     memFolder.Lines.Text:= rsMsgErrNoFiles;
     Exit;


rich2014:
contents of the three files are overlayed.

image

PR #684, opened 2022-11-26
ADD: show link targets in status bar

corvus1:
Useful at times. Without it, the only way I found to see where symlink is pointing is to go open its properties.

### DIFF
--- a/src/fileviews/ucolumnsfileview.pas
+++ b/src/fileviews/ucolumnsfileview.pas
@@ -745,10 +745,13 @@ begin
     if Assigned(AFile) then
       try
         if AFile.IsNameValid then begin
-          if gDirBrackets and (AFile.IsDirectory or AFile.IsLinkToDirectory) then 
-            lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix
-          else 
-            lblInfo.Caption := AFile.Name;
+          lblInfo.Caption := AFile.Name;  
+          if gDirBrackets and AFile.IsDirectory then 
+            lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix;
+          if AFile.IsLink then 
+            lblInfo.Caption := AFile.Name + ' -> ' + AFile.LinkProperty.LinkTo;
+          if gDirBrackets and AFile.IsLinkToDirectory then 
+            lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix + ' -> ' + gFolderPrefix + AFile.LinkProperty.LinkTo + gFolderPostfix;
         end                       
         else 
          if not AInfo then begin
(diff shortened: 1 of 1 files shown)


alexx2000:
It's not optimal code, under Windows all three conditions will be true when link to directory. So it should be like this:
```
if gDirBrackets and AFile.IsLinkToDirectory then
  lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix + ' -> ' + gFolderPrefix + AFile.LinkProperty.LinkTo + gFolderPostfix
else if AFile.IsLink then
  lblInfo.Caption := AFile.Name + ' -> ' + AFile.LinkProperty.LinkTo
else if gDirBrackets and AFile.IsDirectory then
  lblInfo.Caption := gFolderPrefix + AFile.Name + gFolderPostfix
else begin
  lblInfo.Caption := AFile.Name;
end;
```

corvus1:
I'm not set up to test on windows, but I'll take your word for it. :-)

corvus1:
I just realized. If all 3 conditions are true on windows, when it's a link to a directory, the code would actually do the right thing. Purely on accident, of course. :-D It'd treat it as a link to a directory. 

May I ask if we're talking *.lnk files on windows or those ntfs junctions? Or are they treated the same? I'm kinda behind on my windows knowledge. :-)

corvus1:
Anything else I need to address?

alexx2000:
> May I ask if we're talking *.lnk files on windows or those ntfs junctions?

It's about NTFS symlinks and junctions.

> Anything else I need to address?

There are many subtleties to consider. Link property can be `nil`, so code should be more complicated:
```
var
  AText: String

...

if gDirBrackets and AFile.IsLinkToDirectory then
begin
  AText := gFolderPrefix + AFile.Name + gFolderPostfix;
  if Assigned(AFile.LinkProperty) then begin
    AText += ' -> ' + gFolderPrefix + AFile.LinkProperty.LinkTo + gFolderPostfix;
  end;
end
else if AFile.IsLink then
begin
  AText := AFile.Name;
  if Assigned(AFile.LinkProperty) then begin
    AText += ' -> ' + AFile.LinkProperty.LinkTo;
  end;
end
else if gDirBrackets and AFile.IsDirectory then
  AText := gFolderPrefix + AFile.Name + gFolderPostfix
else begin
  AText := AFile.Name;
end;
lblInfo.Caption := AText;
```

corvus1:
That's a good point. Thankfully, it's all wrapped in try..finally, so we wouldn't crash anything, but still.
Fixed.

alexx2000:
`try..finally` block does not process exception. It's only garanties that code from `finally` will be executed. To process exception `try..except` used, [see](https://wiki.freepascal.org/Try).

corvus1:
Oh... I thought they both catch exceptions. That's good to know. :-) Thanks.

PR #703, opened 2022-12-07
UPD: Use TFileViewNotebook.Clear() instead of DestroyAllPages()

rich2014:
the MR about TPageControl.Clear() had been merged in Lazarus Main branch.
https://gitlab.com/freepascal.org/lazarus/lazarus/-/issues/40019

we can remove TFileViewNotebook.DestroyAllPages() and use Clear() in LCL instead.

### DIFF
--- a/src/fmain.pas
+++ b/src/fmain.pas
@@ -5137,7 +5137,7 @@ begin
       begin
         if (DestinationToKeep<>tclLeft) AND (DestinationToKeep<>tclBoth) AND (not(tfadLeft in TabsAlreadyDestroyedFlags)) then
         begin
-          frmMain.LeftTabs.DestroyAllPages;
+          frmMain.LeftTabs.Clear;
           TabsAlreadyDestroyedFlags := TabsAlreadyDestroyedFlags + [tfadLeft]; // To don't delete it twice in case both target are left.
         end;
       end;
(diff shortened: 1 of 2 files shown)


alexx2000:
No, it's too early. I use last Lazarus release to build snapshots. Only macOS builds use Lazarus from `main` branch.

alexx2000:
So I propose this change:
```
uses
  ..., LCLVersion 
....
procedure TFileViewNotebook.DestroyAllPages;
{$if lcl_fullversion >= 2030000}
begin
  Clear;
end;
{$else}
var
  i: Integer;
begin
  for i:=PageCount-1 downto 0 do
    if i<>ActivePageIndex then inherited RemovePage( i );
  inherited RemovePage( 0 );
end;
{$endif}
```

rich2014:
for the codes in the PR:

1. on MacOS, it works fine, we can build with Lazarus Main Branch.

2. on Windows, it is not necessary to build with Lazarus Main Branch. there is no problem with using TPageControl.Clear() in Lazarus 2.2. i did a real test again, and it works fine.

3. the only thing i'm not sure is building with Lazarus 2.2/2.3 on Linux. i did a code review and it should be fine. could you please make a test on Linux?

alexx2000:
3. Linux is most problematic target in this case. It had many random crashes with this function in the past. I will test it.

rich2014:
@alexx2000 

i tested on Linux.

it's too early to use TFileViewNotebook.Clear() instead of DestroyAllPages() with Lazarus 2.3 on Linux.

should we close this PR?

alexx2000:
Yes, I think.

PR #727, opened 2022-12-22
FIX #699: COCOA: cursor in MainControl(Grid) may be always in waiting state

### LINKED ISSUE #699: Hourglass cursor shown indefinitely
### Version

Double Commander
Version: 1.1.0 alpha
Revision: 10491
Commit: 02a004568
Build date: 2022/12/03
Lazarus: 2.3.0.0
FPC: 3.2.2
Platform: x86_64-Darwin-cocoa
OS version: Mac OS X 13.0.1

### Description

Under macOS Ventura, Double Commander sometimes changes the mouse cursor to an hourglass for an "infinite" time (until you go to another directory). However, during that time, DC is actually not busy, but rather responsive. You can execute any action as usual, but with the mouse cursor being an hourglass, which is very irritating.

Steps to reproduce: Select a text file. Then press the F3 key to view it. Then press the Esc key to close the viewer. As a result, the mouse cursor will become an hourglass.

https://user-images.githubusercontent.com/79788982/205487704-a99faa85-8bbf-4ad2-942b-a484d6af36a0.mp4

rich2014:
FIX #699: cursor in MainControl(Grid) may be always in waiting state
1. Cocoa Interface doesn't fully support Cursors, especially for Control-level Cursor.
2. the patch can basically meet DC's needs for Cursor, and will submit patches to Lazarus to fully support more complex needs.

### DIFF
--- a/src/fileviews/ufileviewwithmainctrl.pas
+++ b/src/fileviews/ufileviewwithmainctrl.pas
@@ -36,7 +36,11 @@ uses
   uFileView,
   uDragDropEx,
   uFileViewNotebook,
-  uDebug;
+  uDebug
+{$IFDEF LCLCOCOA}
+  ,uMyDarwin
+{$ENDIF}
+  ;
 
 type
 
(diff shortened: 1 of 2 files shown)


rich2014:
the following screen recording shows the status changes of Cursor.

(sleep(100) is added to TFileSystemListOperation to slow down the reading speed)


https://user-images.githubusercontent.com/102933068/209103506-53ce8049-16bc-4bf0-99c8-7cce7f2219aa.mp4

PR #761, opened 2023-01-18
UPD:  ShortCuts: 11 keys added to uKeyboard for Windows and MacOS

rich2014:
11 keys added to uKeyboard for Windows and MacOS.

this corresponds to the keys that can set ShortCuts on Linux.

tested on MacOS 12 and Windows 11.


zxcat:
Thank you! Works perfect!

JonnyTech:
@rich2014 can this be extended to fix #766? The file mentions VK_RCONTROL and VK_LCONTROL but an Apple keyboard has only one control key so could that be an issue?

rich2014:
i'm not sure what your means.

`ctrl-left` and `ctrl-right` are the

JonnyTech:
Thanks @rich2014, I shall reply in #766

PR #790, opened 2023-02-04
UPD: Format file sizes and counts in file dialogs

avmaksimov:
Now DC uses options for file size when file replacing and also format number of files according to regional settings of OS. Tested only on Windows


avmaksimov:
Hello. I found out that "FloatToStrF(SomeInt, ffNumber, 0, 0)" rounds integer to nearest thousands with big int. So I wrote and recommend to use my new func instead of with huge function calling:

`function IntToStrTS(const AValue: SizeUInt): string;
var i, vSrcLen, vSrcI, vSrcNumberNo, vResLen: byte;
begin
  Str(AValue, Result);
  vSrcLen := Result.Length;

  vResLen := vSrcLen + ((vSrcLen - 1) div 3);
  SetLength(Result, vResLen);

  vSrcI := vResLen;
  vSrcNumberNo := 1;

  for i:= vSrcLen downto 1 do
    begin
      Result[vSrcI] := Result[i];
      Dec(vSrcI);
      if(vSrcNumberNo <> vSrcLen) and (vSrcNumberNo mod 3 = 0) then
        begin
          Result[vSrcI] := FormatSettings.ThousandSeparator;
          Dec(vSrcI);
        end;
      Inc(vSrcNumberNo);
    end;
end;`

So you need just call this function and that's all. What do you think? If you agree with me I could add another commit with it?

Also I hope with function will be added in FPC but there is small chance for it but I'll recommend them.

alexx2000:
Hi. OK, I think we can include this function to `uDCUtils` unit, but `AValue` should be `Int64`.

avmaksimov:
> Hi. OK, I think we can include this function to `uDCUtils` unit, but `AValue` should be `Int64`.

OK I'll make a commit with it later. But is there any reason to use Int64 instead of QWord?

alexx2000:
All standard FreePascal/Lazarus functions which returns file size use it. Also in the arithmetic operations is better to use signed integers to avoid overflow.

avmaksimov:
> All standard FreePascal/Lazarus functions which returns file size use it. Also in the arithmetic operations is better to use signed integers to avoid overflow.

I did it. But what do you think about replacing for ex. "cnvFormatFileSize(FilesPerSecond, uoscNoUnit)" to "IntToStrTS(FilesPerSecond)"? I mean with uoscNoUnit. After this I we can remove uoscNoUnit from set TUsageOfSizeConversion.

P.S. Also I've found another place for the previous task (sync, folder properties in Linux and so on).

avmaksimov:
Please check it out with my comments inside the new commit.

alexx2000:
It is a bad idea to merge in the one commit a large cleanup and functional changes. So split it into two pull requests. Also don't change Image32 component, it is a third party component and it will synchronize with upstream by simple overwrite.

avmaksimov:
@alexx2000 , Is it OK?

avmaksimov:
I can't or I don't know how to split two commits. But it in two different commits.

May be it's because I use Guthub Desktop and choose that I Push all on parent fork.

alexx2000:
No, I don't know. Usually new branch created for each pull request while master synchronized with upstream.
I guess it is simpler save current code, synchronize master with upstream with discard. Create two branches and commit corresponding changes, create new pull requests.

Also I see some clean up are wrong. Some units and variables are used in the platfom specific code (without $IFDEF). So in this case it should be $IFDEF'ed intead remove.

avmaksimov:
> No, I don't know. Usually new branch created for each pull request while master synchronized with upstream.
> I guess it is simpler to save current code, synchronize master with upstream with discard. Create two branches and commit corresponding changes, create new pull requests.
I think I tried this but without success. I couldn't choose a branch or something like it. OK, I'll tried in next time.
BTW is it OK about IntToStrTS? Will you accept it without last commit (with cleaning)?

> Also I see some clean up are wrong. Some units and variables are used in the platform specific code (without $IFDEF). So in this case it should be $IFDEF'ed instead remove.
How can I understand that some module can be surrounded by IFDEF and what IFDEF is needed? About some variables - sorry I see it now.

alexx2000:
> BTW is it OK about IntToStrTS? Will you accept it without last commit (with cleaning)?

Yes

> How can I understand that some module can be surrounded by IFDEF and what IFDEF is needed?

If unit (from which 'uses' clause you want to delete module) does not contain IFDEF's at all then it can be safely cleaned. Otherwise it is needed to compile unit with each IFDEF value.

src/fsyncdirsdlg.pas
```diff
@@ -1820,7 +1820,6 @@ procedure TfrmSyncDirsDlg.SetProgressBytes(AProgressBar: TKASProgressBar;
   CurrentBytes: Int64; TotalBytes: Int64);
 var
   BarText : String;
-  CaptionText : String;
```
alexx2000, line 1823: It is used in the {$IFDEF LCLCOCOA}

src/filesources/gio/ugiofilesourceutil.pas
```diff
@@ -205,10 +205,10 @@ procedure FillAndCount(Files: TFiles; CountDirs: Boolean; out NewFiles: TFiles;
 function FileExistsMessage(SourceFile: TFile; TargetInfo: PGFileInfo; const TargetName: String): String;
```
alexx2000, line 205: This unit does not have `uDCUtils` in the uses clause.

src/udcutils.pas
```diff
@@ -360,7 +366,37 @@ function mbExpandFileName(const sFileName: String): String;
   end;
 end;
 
-function cnvFormatFileSize(iSize: int64; FSF: TFileSizeFormat; Number: integer): string;
+function IntToStrTS(const APositiveValue: Int64): String;
+var i, vSrcLen, vSrcI, vSrcNumberNo, vResLen: byte;
+begin
+  if APositiveValue < 0 then
+    raise Exception.Create('IntToStrTS gets only positive values!');
```
alexx2000, line 373: I think it is better to use `Exit(IntToStr(APositiveValue))` than crash.

src/fsyncdirsdlg.pas
avmaksimov: OK

src/filesources/gio/ugiofilesourceutil.pas
avmaksimov: OK. I'll Add it.

src/udcutils.pas
avmaksimov: I think to better find the error but not to hide. But OK.

avmaksimov:
I managed to create a new pull request for cleaning the code via branches but I'll do it after you commit better because the changes from this commit will be there.

alexx2000:
Don't use `merge` (it creates garbage), always use `rebase` instead. This pull request has incorrect state at this moment. It should have only your changes.

avmaksimov:
OK. I see. I think better if I make backup for two tasks and do this from beginning with removing my fork before.

alexx2000:
Yes, I think it is the easiest and fastest solution.

avmaksimov:
New version: https://github.com/doublecmd/doublecmd/pull/816

PR #804, opened 2023-02-08
ADD: Support parsing of /etc/os-release

wwwutz:
* ADD GetStringsFromEqFile for parsing A=B

* FIX reorder /etc/issue parsing

- Make `src/platform/udcversion.pas` parse `/etc/os-release` which gets setup by almost all systemd based linux distros.
- reorder "last resort" parsing of `/etc/issue` since it may contain anything but system information on most systems

### DIFF
--- a/src/platform/udcversion.pas
+++ b/src/platform/udcversion.pas
@@ -94,7 +94,7 @@ begin
       sl.LoadFromFile(FileName);
       Result := True;
     except
-      on EFilerError do; // Bypass
+      on EStreamError do sl.Free;
     end;
   end;
 end;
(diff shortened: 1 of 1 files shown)


src/platform/udcversion.pas
```diff
@@ -99,6 +99,28 @@ function GetStringsFromFile(FileName: String; out sl: TStringListEx): Boolean;
...
+  sl := nil;
+  if mbFileAccess(FileName, fmOpenRead) then
+  begin
+    sl := TStringListEx.Create;
+    sl.NameValueSeparator:= '=';
+    try
+      sl.LoadFromFile(FileName);
+      Result := True;
+    except
+      on EFilerError do; // Bypass
```
alexx2000, line 118: Here should be:
`on EStreamError do sl.Free;`

wwwutz:
thanks for pointing out.

PR #810, opened 2023-02-13
FIX: ufileview in FlatView mode: issues when there are files with the same name in different subdirectories

rich2014:
FIX: ufileview in FlatView mode: issues when there are files with the same name in different subdirectories:

1. currently only filename is used as HashKey, without path
2. when in normal mode, it works well
3. when in FlatView mode, filename only is not enough to identify a file
4. in the patch, when in FlatView mode, include the path to identify a file

example of related issue:
1. DC in FlatView mode, currently at `currentPath`
2. there is a file in subPath: `currentPath\subPath\file.txt`
3. create file.txt in `currentPath` with another file manager: `currentPath\file.txt`
4. but `currentPath\file.txt` is not shown in DC, and refresh `currentPath\subPath\file.txt` wrongly

### DIFF
--- a/src/fileviews/ufileview.pas
+++ b/src/fileviews/ufileview.pas
@@ -156,6 +156,12 @@ type
     FOnFileListChanged : TOnFileListChanged;
     FLoadingFileListLongTimer: TTimer;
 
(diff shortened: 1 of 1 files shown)


alexx2000:
Replaced by #815

PR #816, opened 2023-02-15
UPD: Format file sizes and counts in file dialogs (v.2)

avmaksimov:
Now DC uses options for file size when file replacing and also format number of files according to regional settings of OS. Tested only on Windows

### DIFF
--- a/src/fFileOpDlg.pas
+++ b/src/fFileOpDlg.pas
@@ -780,8 +780,8 @@ begin
   else begin
     if Operation.State = fsosRunning then ProgressBar.Style := pbstNormal;
     ProgressBar.SetProgress(CurrentFiles, TotalFiles,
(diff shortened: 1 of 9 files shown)


avmaksimov:
I created the new branch in my fork and that's how I solved the problem with two different commit and one pull requests.
May be you add this recommendation to Development Wiki?

Also what about cleaned? I can created it based on master or this fork. Better on this because it constists cleaning after this formatting.

alexx2000:
1. I think it is documented at [GitHub itself](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches). It is a standard practice.

2. I merged pull request, so you can create branch with "cleaning" based on the master.

PR #822, opened 2023-02-18
UPD: cleaning the code from compiler warnings and hints

avmaksimov:
Some of them can avoid future problem. Some of them just cleaning from unused code parts such as units.


avmaksimov:
Uh. It's hard to merge about five times because even after 5 master commits it would be impossible to merge automatically or simply.
So I recommend merge as soon as possible (after checking of course).

PR #825, opened 2023-02-19
UPD: Better looking for command line

avmaksimov:
It's limited by 1/3 of the panel width (simillar the main window width) and after minimize long full path and set width to fit the shorten path.

### DIFF
--- a/src/fmain.lfm
+++ b/src/fmain.lfm
@@ -510,7 +510,6 @@ object frmMain: TfrmMain
         Height = 15
         Top = 6
         Width = 24
-        AutoSize = False
         BorderSpacing.Left = 2
         Caption = 'Path'
         ParentColor = False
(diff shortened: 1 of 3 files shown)


avmaksimov:
Also I added some description to MinimizeFilePath for better understanding and change parameter from MaxLen to MaxWidth that better describes the goal.
I misunderstood before that MaxLen - it's about length of a string (.

alexx2000:
It is better to use  `lblCommandPath.AutoSize = True` as it was before (because height should be autosized). In this case no need to change `lblCommandPath.Width` manually, it will change width automatically.

avmaksimov:
I realized it later)
For the first time the algorithm need to AutoSize = False, but later I decided to use MinimizeFilePath.
So ready.

PR #835, opened 2023-02-26
ADD: Button Help for Options

avmaksimov:
1. Added HelpKeywords for remaining pages.
2. Added Help button to show that you open Help. Button will be invisible when no HelpKeyword.
3. Some small changes: Default and Cancel for TAttributesEdit and change owner for calling from Option page (because when Position=poOwnerFormCenter the owner have to be properly set to avoid showing not at the center as it was before).
4. Remove now duplicate button in Hotlistdir.

Should I commit .po files? It's very difficult to compare. If should than as is or by merging only new important changes?

### DIFF
--- a/src/fattributesedit.lfm
+++ b/src/fattributesedit.lfm
@@ -540,6 +540,7 @@ object frmAttributesEdit: TfrmAttributesEdit
       Width = 94
       Caption = '&OK'
       Constraints.MinHeight = 30
+      Default = True
       Kind = bkOK
       ModalResult = 1
       OnClick = btnOkClick
@@ -550,6 +551,7 @@ object frmAttributesEdit: TfrmAttributesEdit
       Height = 32
       Top = 0
       Width = 118
+      Cancel = True
       Caption = '&Cancel'
       Constraints.MinHeight = 30
       Kind = bkCancel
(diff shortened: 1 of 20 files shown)


avmaksimov:
Oh.. and also I set RowSelect. I think it's more simplier to select a page near the name - not only on text. And looks better.

avmaksimov:
I don't undestand how the button with different function (Help) can be between button for the window. I tried many position (also in caption (with blue background - for ex. "Display" to the right).
About "move it to the left [after filter button](https://imgur.com/a/SkYiM61)" - yes, i tried this version but I don't remember why I refused it). May be because it's too near the filter... But now it looks like better than my one)
P.S. And BTW DC Config window looks more comfortable than TC one).

And what about my question about .po files? Also I think Help button could be on another windows (such as tools or may be in file operation too but as icon). What do you think about it?

avmaksimov:
I tried your version and it is good for me: https://skr.sh/sIb8lFvnmbB?a (commited this version).
Also version with left 4 instead 8: https://skr.sh/sIbiAptOrsU?a

alexx2000:
> Should I commit .po files?

No need. I will regenerate it later.

> Also I think Help button could be on another windows (such as tools or may be in file operation too but as icon). What do you think about it?

F1 is a standard hotkey for help. I think it is not a good idea to add button/icon everywhere. Nobody does that. If we add `HelpKeyword` to other forms then F1 will work there.

PR #841, opened 2023-03-03
UPD: Columns text alignment like values

avmaksimov:
As you can see in Thunderbird. It's more pretty than always left alignment: https://skr.sh/sIf6LJbfTZF


Skif-off:
Maybe optional? It's a bit annoying when switching between column sets.

The simplest way: enable/true by default and without GUI (just describe in `configxml.html` file).

avmaksimov:
It's strange for me what it would be annoying). Do I understand you correctly that you have some column sets with different text layout for the same column and you change column sets often?

Skif-off:
Used for several days and I will stand by my opinion. It seems not a bad feature, looks good, but only if there is no need to actively use different sets of columns (including sets with similar data). Also two or three times I was sure that I had scrolled list to the top of the panel, but I was mistaken :))) I don't know, maybe working with spreadsheets and file managers ruined me?
Is it exists way to make it optional?

avmaksimov:
> Also two or three times I was sure that I had scrolled list to the top of the panel, but I was mistaken :)))

Why? What confused you?

> Is it exists way to make it optional?
Do you mind if I make the option "Column text aligment always to left" only in xml and set to false by default? If there will be many users want it I will add to GUI.

alexx2000:
> Do you mind if I make the option "Column text aligment always to left" only in xml and set to false by default?

Yes, I think. But it should be `True` by default. Total Commander, Windows Explorer always use a left column title alignment.

avmaksimov:
> > Do you mind if I make the option "Column text aligment always to left" only in xml and set to false by default?
> 
> Yes, I think. But it should be `True` by default. Total Commander, Windows Explorer always use a left column title alignment.

OK. May I create visible option? For example: "File Views\Columns"?

I see more modern style in it.

alexx2000:
> May I create visible option, please? For example: "File Views\Columns"?

Yes.

avmaksimov:
Definitely, I'm done with it).

language/doublecmd.ru.po
```diff
@@ -4215,6 +4215,12 @@ msgstr "Фиксированная ширина колонок"
 msgid "Dark mode"
 msgstr "Тёмный режим"
 
+#: tfrmoptionscolumnsview.cbcolumnstitlelikevalues.caption
+#, fuzzy
```
alexx2000, line 4219: Remove `#, fuzzy` and `#| msgid` lines.

src/frames/foptionscolumnsview.lrj
alexx2000, line 1: Fix line-endings.

avmaksimov: I didn't touch this file but make it in Ubuntu. May be I have to save it in Windows?

avmaksimov: I didn't find difference between this file and other lrj-file. Line endings if CRLF.

alexx2000: It is committed wrong. You can see that all lines replaced.

avmaksimov: I see all strings changed again. May be it is github's bug?) 

I don't know how I could commit wrong file what was generated by Lazarus. But I tried again doing small changes.

avmaksimov: I think it's github's bug what shows differences: https://skr.sh/sIvOSJK5taN 

I tried many things but without success. What's why I think so.

alexx2000: Not really, same problem was fixed in the neighbor pull request #856. The problem that Git does not like CRLF line-endings. And sometimes is hard to find what is wrong.

avmaksimov: I tried to commit to LF or CRLF but without success. I don't know may be in commiting something converting line-endings.

decart: You need to set `git config --global core.autocrlf false`. Otherwise, git will automatically convert `CRLF` to `LF` when you commit.

avmaksimov: > You need to set `git config --global core.autocrlf false`. Otherwise, git will automatically convert `CRLF` to `LF` when you commit.

Thank you! It works!

PR #842, opened 2023-03-03
NEW: Icons for Tab from Tab options

avmaksimov:
The tabs gets wider and nicer too.
Also I suggest remove options tb_show_asterisk_for_locked because of now everybody can see status via icon. It's not text file manager anyway).


Skif-off:
You took space away from users (2-3 symbols) and did it with no alternative, right?

Also this feature depends on "Show icons for actions in menus".

avmaksimov:
> You took space away from users (2-3 symbols)
Is it a problem nowadays with resolution from 1300px? Or may be there is shortage of space because of many tabs open (at least more than 10)?
> Also this feature depends on "Show icons for actions in menus"
Yes. I recommend also rename this menu item and set on by default.

JonnyTech:
Please make this an option, I prefer the existing style.

avmaksimov:
> Please make this an option, I prefer the existing style.

Do you mean separate option or may be expand "Show icons for actions in menus" to "Show icons for tabs and actions in menus"?

I think the better the last one.

avmaksimov:
Also I propose to expand the option "Show locked tabs with an asterisk" to "Tabs headers style": "None", "Show locked tabs with an asterisk", "Show tabs icons" (the last one by default).
But I will be grateful if someone help me to expand the old option.

I suggest not to add a new option but use existing as much as possible.

JonnyTech:
> Do you mean separate option or may be expand "Show icons for actions in menus" to "Show icons for tabs and actions in menus"?

Seperate option. Keep existing style as default.

> Is it a problem nowadays with resolution from 1300px? Or may be there is shortage of space because of many tabs open (at least more than 10)?

Please do not assume, I still regularly use 800x600px.

avmaksimov:
> Seperate option

So do you think that somebody wants asterisk for locked and icons?

OK. I'll wait for **Alex2000**'s decision.

alexx2000:
I don't want to mix "Show icons for actions in menus" and show icons in tabs. It should be a two different options. So it is need to create a separate TImageList for tabs.

> Also I propose to expand the option "Show locked tabs with an asterisk" to "Tabs headers style": "None", "Show locked tabs with an asterisk", "Show tabs icons"

Maybe. Does anyone has other ideas?

Skif-off:
Maybe add group with Radio in the end of list, something like:
```
--- Tabs headers style ---------------------
| () Show tabs icons                        |
| () Show title only                        |
|    [] Show locked tabs with an asterisk * |
--------------------------------------------
```
and if the first option is selected, then block "Show locked tabs with an asterisk *" (for the last option, we do not need to correct the translation).

avmaksimov:
> Maybe add group with Radio in the end of list, something like:
> 
> ```
> --- Tabs headers style ---------------------
> | () Show tabs icons                        |
> | () Show title only                        |
> |    [] Show locked tabs with an asterisk * |
> --------------------------------------------
> ```
> 
> and if the first option is selected, then block "Show locked tabs with an asterisk *" (for the last option, we do not need to correct the translation).

I suggest this:
```
--- Tabs headers style -----------------------
| () Show tabs icons                               |
| () Show title only                                  |
| () Show locked tabs with an asterisk * |
-----------------------------------------------------
```
In this way we do not need to correct the translation too.

Skif-off:
@avmaksimov, locked tabs will have their own icon, why do they need an asterisk? And if "Show title only" is selected, we won't have the "locked" marker.

alexx2000:
He offer a three radio buttons, so only one can be selected.

avmaksimov:
Yes. But not to use addition space I think now that combox is preffered option.

Skif-off:
I see three buttons, I don't understand how they should work if only one can be selected. What will the third button do?

1. Title + icons: Normal, PathLocked, PathResets, DirsInNewTab.
2. Title only: Normal, Normal, Normal, Normal.
3. Title only: Normal, Marker, Marker, Marker.

?
Looks strange, because we are offered a choice: headers with icons or headers without icons. The asterisk is (was, is and will be) just a small **_optional_** addition.

More logical:
1. Title + icons: Normal, PathLocked, PathResets, DirsInNewTab.
2. Title only:
- Normal, Normal, Normal, Normal.
- Normal, Marker, Marker, Marker.

avmaksimov:
> I see three buttons, I don't understand how they should work if only one can be selected. What will the third button do?
> 
> 1. Title + icons: Normal, PathLocked, PathResets, DirsInNewTab.
> 2. Title only: Normal, Normal, Normal, Normal.
> 3. Title only: Normal, Marker, Marker, Marker.
> 
> ? Looks strange, because we are offered a choice: headers with icons or headers without icons. The asterisk is (was, is and will be) just a small **_optional_** addition.
> 
> More logical:
> 
> 1. Title + icons: Normal, PathLocked, PathResets, DirsInNewTab.
> 2. Title only:
> 
> * Normal, Normal, Normal, Normal.
> * Normal, Marker, Marker, Marker.

I think it's understandable. The third button: title with asterisk do the title plus asterisk. 

And there will help for it.
BTW every tool needs to be studied))).

And also the space is not robber. Your suggestion needs more place.

Skif-off:
> The third button: title with asterisk do the title plus asterisk.

No :) The first button says "text and icon", the second buttons says "text only", but the third button says about some kind of asterisks only.
The third option only complements and compliments only the second button, so it should look like additional feature.

> studied ... space

I know the measure ;)

avmaksimov:
> > The third button: title with asterisk do the title plus asterisk.
> 
> No :) The first button says "text and icon", the second buttons says "text only", but the third button says about some kind of asterisks only. The third option only complements and compliments only the second button, so it should look like additional feature.

It makes the interface more saficticated. But I understand you and there is a new way from me).
```
Tab header style:
1) Icons;
2) Titles only;
3) Titles only with an asterisk * for locked.
```

I think if we think more about interface so we don't think about saving the old phrase. Somebody needs to translate some new texts anyway.


> > studied ... space
> 
> I know the measure ;)

Rather, it looks like double standards)

Skif-off:
> It makes the interface more saficticated.

More logical choice is a more understandable and friendly way.

Maybe "Titles and icons" in this case? So that no one thinks that only icons will be displayed

> Rather, it looks like double standards)

Vote for me in 2024! :))))
Looks like "take it out of context and say that this a general rule".

avmaksimov:
> > It makes the interface more saficticated.
> 
> More logical choice is a more understandable and friendly way.
> 
> Maybe "Titles and icons" in this case? So that no one thinks that only icons will be displayed
> 
> > Rather, it looks like double standards)
> 
> Vote for me in 2024! :)))) Looks like "take it out of context and say that this a general rule".

I don't mind. So result is:
`Tab header style:
1) Titles and icons;
2) Titles only;
3) Titles only with an asterisk * for locked.`

alexx2000:
https://doublecmd.h1n.ru/viewtopic.php?p=38819#p38819

avmaksimov:
> https://doublecmd.h1n.ru/viewtopic.php?p=38819#p38819
> 
> TC can change a tab icon. I think using icons to mark some tabs can be useful too.

I have answered. I suggest to make in the future).
And in TC it's related to "Design styles". And BTW I think it's too complicated for every user. Better to choose from popup menu but in the future.
I want to finish this pull request properly and I want to finish removing to trash with elevation (((.

avmaksimov:
OK. And I should write it. I found out that the order is more than important in Tabs\Options (gDirTabOptions). I want to remove suboption for locked tabs and add new option "HeaderStyle" but removing is impossible.
So I think there are the following solutions:
1) leave suboption for locked tab for memory))
2) leave suboption for locked tab and add suboption for icons;
3) convert Tabs\Options to different options with value 0 or 1.

I prefer the solution 2 because it's simplier now. But don't judge me strictly))

avmaksimov:
Later I can changes Comboboxes to KASComboboxes to use AutoSize for them.

avmaksimov:
And add Russian language. May be Moldovan too. If the rest of changes is OK.

avmaksimov:
> 2) Titles only;
Hm.. may it's a redundant option? So we can use only checkbox about icons on title and leave default with an asterisk.

Skif-off:
>> Titles only;

> Hm.. may it's a redundant option? So we can use only checkbox about icons on title and leave default with an asterisk.

Why take options away from users? :)

avmaksimov:
> 

Are you sure someone use title without an asterisk? May be its you?)) I don't think so.

JonnyTech:
Please keep existing options, it is not difficult, you risk upsetting somebody somewhere.

avmaksimov:
> Please keep existing options, it is not difficult, you risk upsetting somebody somewhere.

OK. But for me it seems like you thinking somebody who may be doesn't exist).

BTW. I can add later if somebody will ask. So... OK, again. I am waiting for author's answer about code and later I have to make proper TabOrder and localization (or may be he will want to do it by himself).

JonnyTech:
> OK. But for me it seems like you thinking somebody who may be doesn't exist).

Maybe you are correct. Maybe you are not. Either way, certainly add features, but try to avoid removing them.

Skif-off:
> Are you sure someone use title without an asterisk? May be its you?)) I don't think so.

Do you have accurate statistics on the use of program options? Any statistics? :)
You are asked not to break what is. Is this a problem?

> BTW. I can add later if somebody will ask.

I did "will ask" :) I don't look like "somebody"?

avmaksimov:
> You are asked not to break what is. Is this a problem?

No. You can see that I wrote OK before.



> I did "will ask" :) I don't look like "somebody"?

You wrote about "users" before. Do you have such statistics? You didn't wrote that you need it. If you had wrote it before the discussion would stop many messages ago.

Skif-off:
Hmm... GitHub's date formatting is [strikethrough] and [strikethrough]. Was [this post](https://doublecmd.h1n.ru/viewtopic.php?p=38881#p38881) written before or after?

avmaksimov:
> Hmm... GitHub's date formatting is [strikethrough] and [strikethrough]. Was [this post](https://doublecmd.h1n.ru/viewtopic.php?p=38881#p38881) written before or after?

I don't remember. May be the same moment))

avmaksimov:
> https://doublecmd.h1n.ru/viewtopic.php?p=38819#p38819
> 
> TC can change a tab icon. I think using icons to mark some tabs can be useful too.

Could you check and accept this PR please?

meteor1113:
Has this PR been updated?
The current folder tabs are too close together to distinguish easily. Adding icons should help.
Alternatively, any other method that makes tabs easier to tell apart would work too.

alexx2000:
I planned to merge this PR. But macOS does not support icons in the tabs so this PR was frozen.

avmaksimov:
> I planned to merge this PR. But macOS does not support icons in the tabs so this PR was frozen.

May just disable this option/ ability in MacOS?

PR #843, opened 2023-03-04
FIX: Drive list popup for Linux

avmaksimov:
Correct total size for mounted network devices and non mounted [optical] drives.

P.S. Changes in uudev.pas is not necessary but I think it'll bring to error in the future.

May be there is better possibility to fix unmounted optical drive size but I think Linux returns not correct disk size.

### DIFF
--- a/src/udriveslist.pas
+++ b/src/udriveslist.pas
@@ -591,8 +591,7 @@ begin
           Format('%s/%s', [cnvFormatFileSize(FreeSize, uoscHeader),
                            cnvFormatFileSize(TotalSize, uoscHeader)])
       end
-      else if IsAvailable(Drive, False) and
-         (Drive^.DriveSize > 0) then
+      else if (Drive^.DriveSize > 0) then
       begin
         Cells[4, RowNr] := cnvFormatFileSize(Drive^.DriveSize, uoscHeader);
       end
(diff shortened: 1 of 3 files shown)


src/udriveslist.pas
```diff
@@ -591,7 +591,8 @@ procedure TDrivesListPopup.UpdateCells;
           Format('%s/%s', [cnvFormatFileSize(FreeSize, uoscHeader),
                            cnvFormatFileSize(TotalSize, uoscHeader)])
       end
-      else if (Drive^.DriveSize > 0) then
+      else if IsAvailable(Drive, False) and
```
alexx2000, line 594: Bad code, it will break a disk size for unmounted normal disks: HDD, Flash etc.
If `DeviceSize` is wrong for CD disks then simply zero it in the `uUDev` for CD disks.

avmaksimov:
I think this solution is better because are:
1) too complicated to understand for me the right place;
2) the main reason: the check on higher level is much simplier and clearer.

What do you think?

src/udriveslist.pas
avmaksimov: Please see the new fix

alexx2000:
OK.

PR #844, opened 2023-03-04
UPD: Update .gitignore

avmaksimov:
Add main exec file for Linux


avmaksimov:
Also sometimes in Linux I see *.lrj files. Is it needed to add to ignore or sometimes it's useful?

alexx2000:
*.lrj files are useful.

PR #856, opened 2023-03-07
ADD: Config of default context actions

decart:
The default action context menu always shows the default items - Edit and View in the internal or external editor/viewer. This update adds a new configuration item to the "File associations extra" page so that the user can hide these items.

Before:

After:


avmaksimov:
I think better if you rename "Hide default context action" to "Default context actions (View/Edit)" or may be simply "View/Edit" and make default to On.

decart:
Yes, you may be right. It's not a problem to rename it.

src/platform/unix/ushellcontextmenu.pas
```diff
@@ -642,29 +642,33 @@ procedure TShellContextMenu.CreateActionSubMenu(MenuWhereToAdd:TComponent; aFile
   if ContextMenuActionList.Count>0 then
     LocalInsertMenuSeparator;
 
-  // If the external generic viewer is configured, offer it.
-  if gExternalTools[etViewer].Enabled then
+  // If the default context actions not hidden
+  if DefaultContextActions then
```
alexx2000, line 646: It should be `gDefaultContextActions` here and below.

alexx2000:
When default context actions are disabled and file does not have association then submenu is empty. It will be better to not show it in this case.

decart:
Hide the "Actions" submenu if it is empty. In *nix OS, this has already been implemented, I changed the code only for Windows. I used the win32 system function "GetMenuItemCount", which may not be the best solution. Perhaps it would be better to replace the procedure "CreateActionSubMenu" with a function that will return the number of menu items?

src/platform/unix/ushellcontextmenu.pas
decart: Fix it! 👍

src/frames/foptionsfileassocextra.lfm
alexx2000, line 1: Fix line-endings.

decart: Fixed...

alexx2000:
Looks like it did not help. Both *.lfm files had changed line-endings.

decart:
Now it's definitely fixed. For some reason, the file `foptionsfileassocextra.lfm` have different line endings in the `master` branch. Maybe not only in this file...

avmaksimov:
> Now it's definitely fixed. For some reason, the file `foptionsfileassocextra.lfm` have different line endings in the `master` branch. Maybe not only in this file...

I have the same problem with commiting? What's the corect Line-Endings? Only CR as in Linux?

decart:
> I have the same problem with commiting. What's the corect Line-Endings? Only CR as in Linux?

Right now I did as in the `master` branch — CRLF as on Windows. But I can roll back the last commit, and it will be `LF`

decart:
I looked at some files, it looks like there are fifty by fifty values of line endings...😦 How correct is hard to say.

alexx2000:
Looks good now. Line-endings should stay as was before commit.

avmaksimov:
> > I have the same problem with commiting. What's the corect Line-Endings? Only CR as in Linux?
> 
> Right now I did as in the `master` branch — CRLF as on Windows. But I can roll back the last commit, and it will be `LF`

I see that my file in master is CRLF. I tried to commit to LF or CRLF but without success. Could you see my file: src/frames/foptionscolumnsview.lrj ?

decart:
> I see that my file in master is CRLF. I tried to commit to LF or CRLF but without success. Could you see my file: src/frames/foptionscolumnsview.lrj ?

Check `git config --global core.autocrlf`. If the value is `true` or `input`, then git will automatically convert `CRLF` to `LF` on commit.

src/platform/unix/ushellcontextmenu.pas
```diff
@@ -691,7 +695,6 @@ procedure TShellContextMenu.CreateActionSubMenu(MenuWhereToAdd:TComponent; aFile
   // Add shortcut to launch file association cnfiguration screen
   if gIncludeFileAssociation then
   begin
-    LocalInsertMenuSeparator;
```
alexx2000, line 694: Why do you remove this?

decart: Oh, that's my mistake. I got a little confused with the separators. For some reason, I decided to separate all the Extended options items with one delimiter. 😵 I put it back.

PR #859, opened 2023-03-08
UPD: Free space indicator

avmaksimov:
1. New option "Options\File panels\Indicator Threshold Color" that indicate color for non-gradient indicator if free space is less than 10% (constant = 90%; as is in Windows) of total space: 
image

2. For gradient indicator green is expanded from 50% to the same constant 90%.
3. Fake indicator is clickable now so you can see how it will look like.
4. Default fore color is changed to clNavy as more pretty than clBlack.
5. I expand some spaces for comboboxes width for choosing color to fill the largest color name.

P.S. I'm not sure that this constant have to be an option but if you want it I recommend to add it in "Layout" after option "Show free space indicator on drive label".


alexx2000:
Checkbox text overlaps indicator, it was correct before
изображение
изображение

alexx2000:
Indicator overlaps `\` button, last two buttons reversed
изображение

avmaksimov:
Sorry for this garbage). I am only the github rookie)

avmaksimov:
And BTW please consider removing "use" ("Использовать") about "gradient" and it allow to place indicator by left of another components. So I recommend simply: "Gradient indicator" / "Градиентный индикатор".

PR #860, opened 2023-03-09
FIX: Layout in Options: DirHotlist, FilesViews, Icons

avmaksimov:
Also I suggest to move "Dir Hotlist additional" to "Dir Hotlist". I see space after another options.

You can see solved problems:
https://i.imgur.com/vUWF5Q4.png
https://i.imgur.com/2FOHOZK.png
https://i.imgur.com/IDumq3t.png

P.S. It's very hard and boring)))


Skif-off:
> https://i.imgur.com/2FOHOZK.png

You broke the example, try to set `yyyy yyyy yyyy yyyy yyyy.mm.dd hh:mm:ss` and compare.

P.S. Do visual bugs look the same in different themes? "Some functions to select appropriate path" looks fine for me (I am using Greybird which supports GTK2 applications).

avmaksimov:
> You broke the example, try to set `yyyy yyyy yyyy yyyy yyyy.mm.dd hh:mm:ss` and compare.

This incorrect using but I see that all ok either in Windows either in Linux: https://skr.sh/sInqel0n2GG and https://i.imgur.com/R0ZsoBs.png .

So what's the problem? Could add the screenshot please?

avmaksimov:
> I mean that the "Formatting" block is shifted entirely, but before only "Date and time format" was shifted. It doesn't look like a problem (in the vast majority of cases, it would be strange to put a long text there.), just noticed.

Long text is strange too)). But I think it is better than different comboxs widths. As I see DC prefers autowidth but I would set fix size for comboxes and set Align for label with example by left. And also it will look better for high resolution.

But now I just made by analogy.

PR #863, opened 2023-03-11
UPD: jump to Home directory instead of Root directory after unmounting the driver

rich2014:
currently, it would jump to the Root directory after unmounting the driver.

It would be more appropriate to jump to the Home directory.

unix-like system affected only ( eg. linux and MacOS )

### DIFF
--- a/src/platform/unix/ushellcontextmenu.pas
+++ b/src/platform/unix/ushellcontextmenu.pas
@@ -71,11 +71,11 @@ implementation
 uses
   LCLProc, Dialogs, Graphics, uFindEx, uDCUtils, uShowMsg, uFileSystemFileSource,
   uOSUtils, uFileProcs, uShellExecute, uLng, uPixMapManager, uMyUnix, uOSForms,
-  fMain, fFileProperties, DCOSUtils, DCStrUtils, uExts, uArchiveFileSourceUtil
+  fMain, fFileProperties, DCOSUtils, DCStrUtils, uExts, uArchiveFileSourceUtil, uSysFolders
   {$IF DEFINED(DARWIN)}
   , MacOSAll
   {$ELSEIF NOT DEFINED(HAIKU)}
-  , uKeyFile, uMimeActions, uSysFolders
+  , uKeyFile, uMimeActions
     {$IF DEFINED(LINUX)}
   , uRabbitVCS
     {$ENDIF}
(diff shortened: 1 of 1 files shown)


JonnyTech:
Can you make this an option please? This change will break my workflow.

rich2014:
> Can you make this an option please? This change will break my workflow.

hi, Jonny

unix-like system affected only ( eg. linux and MacOS ).

in these systems, what scenario would it be inconvenient to jump to the home directory?

JonnyTech:
Hey Rich, I do work mostly on Linux and MacOS systems, and have mounts in various places, grouped by projects, outside of my home folder. What I would like, if possible, is to make the folder as a configurable option, that way I can set it per project/session and it would save me a keystoke, but otherwise drive root please.

rich2014:
@JonnyTech 

1. i understand and respect that everyone has different habits
2. however, in a Unix-like system, being in the root directory is dangerous and not recommended, and 99% of the cases should not be in the root directory
3. so i think we should try to avoid setting the current path to the root directory. especially in the case of unmounting the drive, it is not the user who switches to the root directory, but DC automatically jumps to the root directory, which should be avoided
4. for the other 1% of the cases, it should be the user actively switching to the root directory
5. it's better not to add more configure options in this case, because the current DC configure options are already large enough and complex enough

JonnyTech:
@rich2014 - yes, I understand your comments, but I am not in the root directory, instead I may be in something like `/projects/project_name` or `/home/projects/project_name` but not in `/home/$USER/projects/project_name`, hence my request to either:

> Can you make this an option please?

or

> make the folder as a configurable option

But that was just a request, if not possible or too difficult then I can live with the change. Sorry to be a nuisance. Thank you :)

rich2014:
@JonnyTech 

maybe Hohtlist is more suitable for you.

after all, it is tricky to rely on unmounting driver.

JonnyTech:
> maybe Hohtlist is more suitable for you.

Yes, thank you, I already use that often, appreciate the guidance anyway.

rich2014:
@alexx2000 

could you please check and judge this PR?

rich2014:
@alexx2000

rebased, and tested on MacOS and Linux.

could you please check and judge this PR?

alexx2000:
I don't know how is better. This behavior is not important for me personally.

However I agree that home directory is better than root directory.

What about third variant: go to mount point parent directory?

PR #868, opened 2023-03-15
FIX: file icon issue in FileProperties Form

rich2014:
FIX: file icon issue in FileProperties Form.

1. display file icon correctly

2. on Retina screen, display larger sized file icon (up to 48x48)

tested on MacOS 12 and Linux.


rich2014:
image

PR #878, opened 2023-03-19
ADD: Elevation delete to trash

avmaksimov:
It tested in Windows 10 and Ubuntu 22.04. I had to separate some functions to uURIHandling and other changes for Linux.

Also I changed default behavior for Windows: it allow to trash only if a registry value equal some value and default not - it's right for network drives for example.

Also I want to add checking on Windows for size of removing files. Because DC could show information about impossibility to trash files more than max size for trash on this disk.

### DIFF
--- a/components/doublecmd/dcunix.pas
+++ b/components/doublecmd/dcunix.pas
@@ -118,6 +118,15 @@ type
   TGroupRecord = group;
   PGroupRecord = ^TGroupRecord;
 
+{$IF DEFINED(UNIX)}
+{en
+   Find mount point of file system where file is located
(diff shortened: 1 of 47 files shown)


avmaksimov:
Unfortunately I found some small mistakes and have to fix it.

avmaksimov:
I hope that's all. But it's needed to do some work with checking file or folder before deleting in Windows for max file size. So please check it.

And also I suggest to do this checking in TFileSystemDeleteOperation.ProcessFile . Do you agree? mbCheckTrash checks only for drive.

PR #880, opened 2023-03-20
FIX: key ESC/RETURN handled in Options Dialog & KasPathEdit

rich2014:
fix key ESC/RETURN handled in KasPathEdit & Options Dialog:

1. after the Key ESC/RETURN handled in TKASPathEdit.handleSpecialKeys(), Key should be eaten (there is a bug in #876)

2. in fOptions, set `btnCancel.Cancel`=`true` , and remove `alOptionsActionList`, handle `ESCAPE` correctly on all platforms

3. tested on Windows 11, MacOS12 and Linux

### DIFF
--- a/src/foptions.pas
+++ b/src/foptions.pas
@@ -63,8 +63,6 @@ type
     TreeFilterEdit: TTreeFilterEdit;
     tvTreeView: TTreeView;
     splOptionsSplitter: TSplitter;
-    alOptionsActionList: TActionList;
-    actCloseWithEscape: TAction;
     procedure btnCancelClick(Sender: TObject);
     procedure btnHelpClick(Sender: TObject);
     procedure FormClose(Sender: TObject; var CloseAction: TCloseAction);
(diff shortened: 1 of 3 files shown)


src/foptions.pas
```diff
@@ -222,7 +219,7 @@ procedure TfrmOptions.FormClose(Sender: TObject; var CloseAction: TCloseAction);
 
 procedure TfrmOptions.FormCloseQuery(Sender: TObject; var CanClose: boolean);
 begin
-  CanClose := (ModalResult in [mrOK, mrCancel]) or CycleThroughOptionEditors(False);
```
alexx2000, line 225: Why do you remove `mrCancel`? DC should not ask user to save changes when user press cancel. It should ask only when user close window by "X" title button.

rich2014: for careless closing the Form.

press Cancel with quering, it seems a bit risky.

alexx2000: It was such way five years ago. It is very annoying, I press `Cancel` when I want to cancel all changes. Why DC should ask me again about this?

avmaksimov: > It was such way five years ago. It is very annoying, I press `Cancel` when I want to cancel all changes. Why DC should ask me again about this?

I think it's normal situation if DC will asks about it too. You could by mistake press another button (.

rich2014:
@alexx2000

PR #883, opened 2023-03-23
Fix: Sorting after first install and small refactoring

avmaksimov:
Subj and I don't understand for what exists DoSortByFunctions. I think for the future so what's why I decided to leave this function,

But all callings with one element array and this cycle doesn't make sense:

```
// Check if there is already sorting by one of the functions.
  // If it is then reverse direction of sorting.
  for i := 0 to Length(FileFunctions) - 1 do
  begin
    SortDirection := GetSortDirection(CurrentSorting, FileFunctions[i]);
    if SortDirection <> sdNone then
    begin
      SortDirection := ReverseSortDirection(SortDirection);
      Break;
    end;
  end; 
```
So ... to live or to die this function and if to die I can merge DoSortByFunctions with my new one DoSortOneFunction.


avmaksimov:
The problem in Russian: https://doublecmd.h1n.ru/viewtopic.php?t=7656

avmaksimov:
I removed my DoSortOneFunction after some optimizations.

PR #886, opened 2023-03-25
UPD: adding padding for file list in FileView to the bottom and to the right for column right text alignment

avmaksimov:
1. Adding padding for right alignment columns.
2. Adding padding for the last row in file list (grid).

Before: https://skr.sh/sJ1gTQUPLLl
After: https://skr.sh/sJ16s6rVxyG

Sorry for colours. It's the best for testing purpose.

### DIFF
--- a/src/fileviews/ucolumnsfileview.pas
+++ b/src/fileviews/ucolumnsfileview.pas
@@ -507,7 +507,7 @@ begin
   if gInplaceRenameButton and (ARect.Right + edtRename.ButtonWidth < dgPanel.ClientWidth) then
     Inc(ARect.Right, edtRename.ButtonWidth);
 
-  edtRename.SetBounds(ARect.Left, ARect.Top, ARect.Right - ARect.Left, ARect.Bottom - ARect.Top);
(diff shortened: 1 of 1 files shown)


alexx2000:
1. You can use `aRect.Height` instead `aRect.Bottom - aRect.Top`.
2. [Here](https://github.com/doublecmd/doublecmd/blob/f70bf1b9efb91325009c5b79f445af346057b65d/src/fileviews/ucolumnsfileview.pas#L650) `aRect.Height` should be used too instead `dgPanel.RowHeights[FileIndex]`.

avmaksimov:
> 1. You can use `aRect.Height` instead `aRect.Bottom - aRect.Top`.
> 2. [Here](https://github.com/doublecmd/doublecmd/blob/f70bf1b9efb91325009c5b79f445af346057b65d/src/fileviews/ucolumnsfileview.pas#L650) `Result.Height` should be used too instead `dgPanel.RowHeights[FileIndex]`.

Ready. Also I've found some similar places in the unit.

Please check it out.

PR #892, opened 2023-03-29
Updated Slovak language

dodog:


language/doublecmd.sk.mo
alexx2000, line 1: This file is not needed.

language/lcl/lclstrconsts.sk.po
alexx2000, line 1: https://github.com/doublecmd/doublecmd/blob/master/language/lcl/README.txt

PR #921, opened 2023-04-06
UPD: DARWIN: unmount driver

rich2014:
use `NSWorkspace.sharedWorkspace.unmountAndEjectDeviceAtPath()` instead of `diskutil unmount`.

`diskutil unmount` has a high probability of not successfully unmounting.

### DIFF
--- a/src/platform/unix/darwin/umydarwin.pas
+++ b/src/platform/unix/darwin/umydarwin.pas
@@ -3,7 +3,7 @@
    -------------------------------------------------------------------------
    This unit contains specific DARWIN functions.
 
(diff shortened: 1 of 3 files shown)


alexx2000:
This patch use same function for both commands (unmount/eject). But drive button context menu items called "Unmount" and "Eject". So it this case menu should have only one item "Unmount and eject".

rich2014:
@alexx2000  

fixed.

although `unmount` can be implemented separately, but for end users on MacOS, unmount alone is easy to cause trouble, and cannot be remounted through DC or Finder.

therefore, the single menu of unmount/eject is selected in ContextMenu after comparison.

2. unmount

Issue #60, opened 2021-03-16
The creation time is not preserved


ghost:
It does not preserve the creation time for directories and files on the Double Commander macOS version. It always sets on the transfer time.
[doublecmd-0.9.10-9640.qt.x86_64.dmg](https://sourceforge.net/projects/doublecmd/files/DC%20for%20Mac%20OS%20X%2064%20bit/Double%20Commander%200.9.10%20beta/doublecmd-0.9.10-9640.qt.x86_64.dmg/download)

When I test the Double Commander Windows version then there are no problems by default it preserves the creation time when copying directories and files locally.
[doublecmd-0.9.10.x86_64-win64.exe](https://sourceforge.net/projects/doublecmd/files/DC%20for%20Windows%2064%20bit/Double%20Commander%200.9.10%20beta/doublecmd-0.9.10.x86_64-win64.exe/download)

alexx2000:
This function is not supported in macOS version.

Guntriaske457:
> This function is not supported in macOS version.

Thanks for the infomation.

Is it possible to support it? I have many files and directories on my macOS 11 MacBook and would like to preserve all 3 timestamps (created, modified and accessed time) when copying. So just like the Windows version.

alexx2000:
Yes, this can be added. But I don't sure that I will be able to continue support macOS at all.

Guntriaske457:
> Yes, this can be added. But I don't sure that I will be able to continue support macOS at all.

It would be good if you continue to support macOS (Unix like), because it is one of the few operating systems besides Windows that fully support created time.

rich2014:
I test on MacOS 12.6, it preserve create time on APFS.

has this problem been fixed since then?

alexx2000:
It is impossible. DC does not copy creation time under Unix-like operating systems, [see](https://github.com/doublecmd/doublecmd/blob/0611623c6a266e1116f3682327d84d99d78aadf0/components/doublecmd/dcosutils.pas#L542). DC copy only access and modification times.

PR #932, opened 2023-04-12
ADD: copy file with CreationTime on MacOS

rich2014:
see #60


alexx2000:
Looks like structure StatInfo under x86_64 does not have a "st_birthtime" field.

rich2014:
> Looks like structure StatInfo under x86_64 does not have a "st_birthtime" field.

#935 temp patch for x86_64

PR #955, opened 2023-04-28
ADD: FileTime: at least millisecond precision supported on Linux and MacOS

rich2014:
### currently DC only supports FileTime with second precision on Linux and MacOS:

1. FileTime includes ModificationTime, CreationTime, AccessTime, ChangeTime

2. when copying files in DC, the fractional parts of FileTime are ignored

3. in SetFileProperties Dialog, the milliseconds are ignored

### this PR adds support for at least millisecond precision, which is usually required, as on Windows. the ideas are:

1. because it involves modifications in many places, the principle of this PR is to make as few changes as possible, so the original function and the `Ex` function version are retained at the same time. after this PR is merged, it should take one or two more PRs to replace them all with the new `Ex` version.

2. record `TFileTimeEx` added, which corresponds to `TFileTime`, but adds support for nanosecond precision. not directly modifying `TFileTime` is the reason mentioned in 1st, to minimize the places that need to be modified.

3. the `Ex` series functions added:

`in DCBasicTypes`
function FileTimeToDateTimeEx(FileTime: TFileTimeEx) : TDateTime;

`in DCDateTimeUtils`
function DateTimeToFileTimeEx(DateTime : TDateTime): TFileTimeEx;
function UnixFileTimeToDateTimeEx(UnixTime: TFileTimeEx): TDateTime;
function DateTimeToUnixFileTimeEx(DateTime: TDateTime): TFileTimeEx;

`in DCOSUtils`
function mbFileSetTimeEx(const FileName: String; ModificationTime: TFileTimeEx;  CreationTime: TFileTimeEx; LastAccessTime  : TFileTimeEx): Boolean;
function mbFileGetTime(const FileName: String; var ModificationTime: TFileTimeEx; var CreationTime: TFileTimeEx; var LastAccessTime: TFileTimeEx): Boolean;
(mbFileGetTime() extend to support nanosecond precision)

`in uAdministrator`
function FileSetTimeExUAC(const FileName: String; ModificationTime: TFileTimeEx; CreationTime: TFileTimeEx; LastAccessTime: TFileTimeEx): LongBool;

4. modifications in the application layer

`in uFileSystemFileSource`
function TFileSystemFileSource.CreateFile(const APath: String; pSearchRecord: PSearchRecEx): 

`in uFileSystemSetFilePropertyOperation`
function TFileSystemSetFilePropertyOperation.SetNewProperty(aFile: TFile; aTemplateProperty: TFileProperty): TSetFilePropertyResult;

### fully tested on Linux, MacOS 12 and Windows 11.

### DIFF
--- a/components/doublecmd/dcbasictypes.pas
+++ b/components/doublecmd/dcbasictypes.pas
@@ -22,6 +22,9 @@
 
 unit DCBasicTypes;
 
+{$mode objfpc}{$H+}
+{$modeswitch advancedrecords}
+
 interface
 
 type
@@ -35,6 +38,7 @@ type
 
 {$IFDEF MSWINDOWS}
   TFileTime = TWinFileTime;
+  TFileTimeEx = TFileTime;
 {$ELSE}
   // Unix time (UTC).
   // Unix defines time_t as signed integer,
@@ -44,6 +48,15 @@ type
   {$ELSE}
   TFileTime = DWord;
   {$ENDIF}
+
+  TFileTimeEx = record
+    public
+      sec: int64;
+      nanosec: int64;
+    public
+      constructor create( aSec:int64; aNanosec:int64=0 );
+      class operator =(l,r : TFileTimeEx): Boolean;
+  end;
 {$ENDIF}
 
   TUnixFileTime = TFileTime;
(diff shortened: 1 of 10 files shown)


rich2014:
image

Skif-off:
Forgive my curiosity (and ignorance :)), what benefits does it provide?

P.S. As I understand, DC uses seconds with Lua too.

rich2014:
@Skif-off 

it is useful when you need to preserve file times precisely.

for example, we have a directory with many small files that are continuously generated. if the precise time is not preserved, the time of a large number of files may become the same after copying. the order will change when sorting by time.

and, `DC on Windows` already supports precise time.

and, `SetFileProperties Dialog` already supports precise time inputting.

alexx2000:
Also we have a related [bug report](https://doublecmd.sourceforge.io/mantisbt/view.php?id=2286).

PR #956, opened 2023-04-29
dcpcrypt2.pas : upstream hash fix for large files

wesinator:
update the dcpcrypt2.pas code to v2.0.4.2
https://gitlab.com/freepascal.org/lazarus/ccr/-/issues/31934


alexx2000:
But DC does not use this method.

PR #967, opened 2023-05-03
FIX #963: MacOS crash in fswatch

### LINKED ISSUE #963: Crash on macos-smbfs browsing
### Version

Double Commander
Version: 1.1.0 alpha
Revision: 10714
Commit: 51ca918fa
Build date: 2023/05/01
Lazarus: 2.3.0.0
FPC: 3.2.2
OS version: Mac OS X 13.3.1

### Description

Latest snapshot bring crash (not handled Access violation) when browsing some directories mounted with smbfs
How to reproduce - does not know, because not all dirs crashing, but always easy to reproduce for me.
Just walk through some dirs mounted with smbfs
Snapshot 10710 does not have this crash

[dc-crash.txt](https://github.com/doublecmd/doublecmd/files/11367275/dc-crash.txt)
[dc-crash2.txt](https://github.com/doublecmd/doublecmd/files/11367276/dc-crash2.txt)

exidler:
_watchPaths and _watchRealPaths have non atomic changes => better use full synchronisation,
but this changes mostly avoiding out of range in loop

### DIFF
--- a/src/platform/unix/darwin/udarwinfswatch.pas
+++ b/src/platform/unix/darwin/udarwinfswatch.pas
@@ -42,7 +42,7 @@ interface
 
 uses
   Classes, SysUtils, Contnrs, SyncObjs,
-  MacOSAll, CocoaAll, BaseUnix;
+  MacOSAll, CocoaAll, BaseUnix, Math;
 
 type TDarwinFSWatchEventCategory = (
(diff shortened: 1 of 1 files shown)


rich2014:
@exidler 

thx for your pointing.

I'll add a synchronisation..

rich2014:
@alexx2000 

hi, Alex, we should merged #968 instead of #967.

#967 is a temp fix.

PR #977, opened 2023-05-07
FIX: memory leaks in udarwinfswatch

rich2014:
In Cocoa, as a created and resident thread, NSAutoreleasePool needs to be added to release memory for the Autorelease variable, otherwise it will not be released until the thread ends.

### DIFF
--- a/src/platform/unix/darwin/udarwinfswatch.pas
+++ b/src/platform/unix/darwin/udarwinfswatch.pas
@@ -534,11 +534,14 @@ procedure cdeclFSEventsCallback(
   eventFlags: FSEventStreamEventFlagsPtr;
   {%H-}eventIds: FSEventStreamEventIdPtr ); cdecl;
 var
+  pool: NSAutoReleasePool;
   watcher: TDarwinFSWatcher absolute clientCallBackInfo;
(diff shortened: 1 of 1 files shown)


rich2014:
in Cocoa, as a created and resident thread, NSAutoreleasePool needs to be added to release memory for the Autorelease variable, otherwise it will not be released until the thread ends.

rich2014:
@alexx2000

hi, Alex, are there similar resident threads in DC?

alexx2000:
Hi, I cannot say exactly.

rich2014:
why? 

i observe that there always seem to be many threads when the DC is running.

if some threads do not end for a long time, then it is a similar situation.

alexx2000:
Because I am not remember all. Also I don't sure that someone thread use Cocoa objects.

alexx2000:
If thread does not use Autorelease variables then nothing to free. Is it right?

rich2014:
in theory, as long as it is a resident thread, it should be wrapped with AutoReleasePool.

because even if we don't explicitly use autorelease ourselves, we may call methods that use autorelease.

for example, in StringToNSString()

PR #1060, opened 2023-07-27
FIX: shell execute parameters

TheNicker:
Extend the capability of cm_ShellExecute to take parameters.
Now cm_ShellExecute can be used to run external programs **with** parameters

### DIFF
--- a/src/umaincommands.pas
+++ b/src/umaincommands.pas
@@ -1404,7 +1404,7 @@ var
 begin
   if Length(Params) > 0 then
     with frmMain do
-    ShellExecute(PrepareParameter(Params[0]))
+    ExecCmdFork(PrepareParameter(Params[0]))
   else
     with frmMain.ActiveFrame do
     begin


alexx2000:
cm_ShellExecute opens file/folder in the associated program. So proposed change are wrong.

TheNicker:
I was trying to be as less as intrusive as possible, yeap it isn't perfect.

Do we want to keep the responsibility of cm_ShellExecute unchanged, or not ? 
With this proposal, it can do two things, and it achieves the desired functionality (running anything)
If left unchanged then can another command can be introduced ?

alexx2000:
cm_ShellExecute should be unchanged because proposed patch breaks command logic. 

> If left unchanged then can another command can be introduced ?

DC can execute external commands directly. Why do you need additional internal command?

TheNicker:
Other than using the toolbar as shown below, I'm not aware of any way to run external command in DC
image

I would like to declare external command directly and bind it to a hotkey.

alexx2000:
In this case a new command (like `cm_ExecuteCommand`) should be introduced.

PR #1061, opened 2023-07-27
Fix: directory change to non-normalized paths - issue #130

### LINKED ISSUE #130: Add Support for non-normalized file paths when assigning new directory path

TheNicker:
Fixes #130 
Sometimes when using source code from different sources, non-normalized paths can be a pain and currently double commander can't read those.

A perfect cross platform solution would be the logic in the function `path::lexically_normal()`  found in the following link:
https://github.com/gcc-mirror/gcc/blob/b24acae8f4d315a5b071ffc2574ce91c7a0800ca/libstdc%2B%2B-v3/src/c%2B%2B17/fs_path.cc#L1686

However, in the meantime I introduce this solution:
The WINAPI `GetFullPathName` function is superior to the current double commander implementation, it solves complex cases of non-normalized path forms such as `c:/windows\/system32///\..`

### DIFF
--- a/components/doublecmd/dcstrutils.pas
+++ b/components/doublecmd/dcstrutils.pas
@@ -375,11 +375,16 @@ const
   AllowPathDelimiters : set of char = ['\','/'];
 var
   I : LongInt;
+  uriPos : Integer;
 begin
   Result:= Path;
   // If path is not URI
-  if Pos('://', Result) = 0 then
-  begin
+  uriPos := Pos('://', Result);
+  if (uriPos = 0)
+{$IF DEFINED(MSWINDOWS)}
+     or ( (uriPos = 2) and  (Path[1] in ['A'..'z']) )
+{$ENDIF} then
+ begin
     for I:= 1 to Length(Path) do
       if Path[I] in AllowPathDelimiters then
         Result[I]:= DirectorySeparator;
--- a/src/udcutils.pas
+++ b/src/udcutils.pas
@@ -358,13 +358,19 @@ end;
 
 function mbExpandFileName(const sFileName: String): String;
 begin
-  if (Pos('://', sFileName) > 0) then
+  if (Pos('://', sFileName) > 2) then
     Result:= sFileName
   else begin
     Result:= NormalizePathDelimiters(sFileName);
     Result:= ReplaceEnvVars(Result);
     if Pos(PathDelim, Result) <> 0 then
       Result:= ExpandFileName(Result);
+
+{$IF DEFINED(MSWINDOWS)}
+// Remove double backslash '\\' after calling 'ExpandFileName'
+    if (Pos(':\\', Result) = 2) and (Result[1] in ['A'..'z'])  then
+      Result:= Result.Remove(2, 1);
+{$ENDIF}
   end;
 end;
 


Skif-off:
Why `GetFullPathNameA ` and `PChar(Path)`? It's not Unicode.
Why not `GetFullPathNameW` and `PWideChar(Path)`? Two examples [here](https://github.com/doublecmd/doublecmd/blob/ab64636e64f0712238c423104e080dc4bed4d0af/components/doublecmd/dcntfslinks.pas).

`NormalizePathDelimiters`, `mbExpandFileName `, `ExpandAbsolutePath`?

TheNicker:
> Why `GetFullPathNameA ` and `PChar(Path)`? It's not Unicode. Why not `GetFullPathNameW` and `PWideChar(Path)`? Two examples [here](https://github.com/doublecmd/doublecmd/blob/ab64636e64f0712238c423104e080dc4bed4d0af/components/doublecmd/dcntfslinks.pas).
> 
> `NormalizePathDelimiters`, `mbExpandFileName `, `ExpandAbsolutePath`?

`ExpandAbsolutePath` receives String as a parameter, String in double commander is AnsiString.

@Skif-off , Thanks to you I've rechecked the change and now setting the active file name is also supported. e.g: 
`c:/windows\/system32///\../explorer.exe`

Path normalization happens sooner, closer to the user input event.

Pull request updated, please re-review

Skif-off:
> String in double commander is AnsiString.

As I understand, AnsiString is a sequence of bytes that can contain bytes of characters in UTF-8 encoding. (I thunk, otherwise, DC will not be able to work with non-ASCII characters.)

alexx2000:
> ExpandAbsolutePath receives String as a parameter, String in double commander is AnsiString.

No, `AnsiString` really are `UTF8String`. So `GetFullPathNameW` should be used and it should be called like:
```
var
  dwSize: DWORD;
  P: PWideChar = nil;
  wsResult: UnicodeString;
begin
  SetLength(wsResult, MaxSmallInt + 1);
  dwSize:= GetFullPathNameW(PWideChar(UTF16LongName(FileName)), MaxSmallInt, PWideChar(wsResult), P);
  if (dwSize = 0) or (dwSize > MaxSmallInt) then
    Result:= FileName
  else begin
    SetLength(wsResult, dwSize - 1);
    Result:= CeUtf16ToUtf8(wsResult);
  end; 
```
Also I don't sure that `ExpandAbsolutePath` should be patched because it also used with virtual paths. This is much more complicated than it seems. I guess it is better to change [mbExpandFileName](https://github.com/doublecmd/doublecmd/blob/ab64636e64f0712238c423104e080dc4bed4d0af/src/udcutils.pas#L359) function instead (and use it).

TheNicker:
Indeed `String` is UTF8, thanks for pointing that out.
mbExpandFileName is a good idea and can be almost used as is, it solve every case expect multiple directory separators after the drive path , e.g c:///.


I've updated the PR and added [mbExpandFileNameForLocalFile](https://github.com/doublecmd/doublecmd/blob/6806b89cb0a82148a9e93d6618c476d864605a64/src/udcutils.pas#L367)

`ExpandAbsolutePath` is untouched

alexx2000:
No need a new function. Change and use `mbExpandFileName`.

src/fileviews/ufileviewheader.pas
```diff
@@ -232,7 +232,7 @@ procedure TFileViewHeader.onKeyRETURN(Sender: TObject);
   NewPath: String;
   AClass: TFileSourceClass;
 begin
-  NewPath:= NormalizePathDelimiters(FPathEdit.Text);
+  NewPath:= mbExpandFileNameForLocalFile(FPathEdit.Text);
   NewPath:= ReplaceEnvVars(ReplaceTilde(NewPath));
```
alexx2000, line 236: Remove `ReplaceEnvVars` from here, it is already called in the `mbExpandFileName`.

TheNicker:
> No need a new function. Change and use `mbExpandFileName`.

mbExpandFileName doesn't solve the case of two directory separators after the drive name, e.g. c://

alexx2000:
Yes, we can't fix `c://` because it looks like URL, only `c:\\`. URL can be entered as path so we can't remove URL checking.
```
function mbExpandFileName(const sFileName: String): String;
begin
  if (Pos('://', sFileName) > 0) then
    Result:= sFileName
  else begin
    Result:= NormalizePathDelimiters(sFileName);
    Result:= ReplaceEnvVars(Result);
    if Pos(PathDelim, Result) <> 0 then
      Result:= ExpandFileName(Result);
{$IF DEFINED(MSWINDOWS)}
    if (Result.Length > 3) and (Result[3] = PathDelim) and (Result[4] = PathDelim) then
      Result:= Result.Remove(2, 1);
{$ENDIF}
  end;
end;
```

TheNicker:
`c://` need to be fixed, it's legal in Windows and Linux so it supposed to be legal in Double commander.
The only thing that isn't legal is backslash in Linux

TheNicker:
'c://' need to be fixed. It's legal in Windows so it should be legal in double commander.
If `NormalizePathDelimiters` is adjusted for Windows drives then we can just remove two backslahes in Windows
Assuming that there are no single scheme URI's e.g. in the form [x]:// 
 and the solution might be:
```
function mbExpandFileName(const sFileName: String): String;
begin
  if (Pos('://', sFileName) > 2) then
    Result:= sFileName
  else begin
    Result:= NormalizePathDelimiters(sFileName);
    Result:= ReplaceEnvVars(Result);
    if Pos(PathDelim, Result) <> 0 then
      Result:= ExpandFileName(Result);

{$IF DEFINED(MSWINDOWS)}
// Remove double backslash '\\' after calling 'ExpandFileName'
    if (Pos(':\\', Result) = 2) and (Result[1] in ['A'..'z'])  then
      Result:= Result.Remove(2, 1);
{$ENDIF}
  end;
end;
```
```
function NormalizePathDelimiters(const Path: String): String;
{$IFDEF UNIX}
begin
  Result:= Path;
end;
{$ELSE}
const
  AllowPathDelimiters : set of char = ['\','/'];
var
  I : LongInt;
  uriPos : Integer;
begin
  Result:= Path;
  // If path is not URI
  uriPos := Pos('://', Result);
  if (uriPos = 0)
{$IF DEFINED(MSWINDOWS)}
     or ( (uriPos = 2) and  (Path[1] in ['A'..'z']) )
{$ENDIF} then
 begin
    for I:= 1 to Length(Path) do
      if Path[I] in AllowPathDelimiters then
        Result[I]:= DirectorySeparator;
  end;
end;
{$ENDIF}                                                               
```

TheNicker:
Any comments on the above, Before I update the PR?

alexx2000:
I don't sure about `NormalizePathDelimiters`. This change do nothing in this case.

TheNicker:
> I don't sure about NormalizePathDelimiters. This change do nothing in this case.

`mbExpandFileName` expects path delimiters to be normalized. 

`NormalizePathDelimiters`  is changed to distinguish between a system drive path and a URI, otherwise it won't normalize path delimiters and `mbExpandFileName` won't work as desired.

In this PR `NormalizePathDelimiters` creates exception just for windows where `[x]://` is now considered as a system drive and not a URI, assuming x is a letter from 'A' to 'z'.

PR updated.

PR #1217, opened 2023-08-10
Update changelog.txt

JonnyTech:
Taken from sourceforge mantis changelog page

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -1,5 +1,1065 @@
-﻿--------------------------------------------------------------------------------
+Double Commander - Change Log
 https://doublecmd.sourceforge.io/mantisbt/changelog_page.php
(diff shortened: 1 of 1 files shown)


Skif-off:
What about [Changes in version N](https://github.com/doublecmd/doublecmd/wiki)?
`git log`?

JonnyTech:
Whatever works. Decision for @alexx2000. I was just tired of seeing an old document bundled with each new release.

Skif-off:
Sorry, I don't understand why an _incomplete_ list is needed.

If this is a question, then there are lists of changes on the DC Wiki (the link is in the FAQ and in the header of every help page): just copy and save.
(If anyone needs specific details, there is the repository, the old repository (switched to read-only state) and the old bug tracker (switched to read-only state).)

PR #1259, opened 2023-09-09
Fix: missing 'mbExpandFileName' invocations for 273759d1 for issue #130

TheNicker:


TheNicker:
Apparently we forgot to invoke  'mbExpandFileName'. 
Use cases from #130 seems to work now

src/fileviews/ufileviewheader.pas
```diff
@@ -234,8 +234,8 @@ procedure TFileViewHeader.onKeyRETURN(Sender: TObject);
   NewPath: String;
   AClass: TFileSourceClass;
 begin
-  NewPath:= NormalizePathDelimiters(FPathEdit.Text);
   NewPath:= ReplaceEnvVars(ReplaceTilde(NewPath));
```
alexx2000, line 237: `FPathEdit.Text` and `NewPath` are confused here.

TheNicker: It's on purpose. 
it's the same as in [fmain.pas:5903](https://github.com/doublecmd/doublecmd/pull/1259/files#diff-7b416c83e34c76fdacbb6e2854de3c561dd8177c9401896c5e8df0677671d1cbR5903) and [fmain.pas:5910](https://github.com/doublecmd/doublecmd/pull/1259/files#diff-7b416c83e34c76fdacbb6e2854de3c561dd8177c9401896c5e8df0677671d1cbR5910)

Shouldn't it be the same for both code paths ?

alexx2000: Я не про порядок вызова функций, сейчас выходит такой код:
```
var
  NewPath: String;
  AClass: TFileSourceClass;
begin
  NewPath:= ReplaceEnvVars(ReplaceTilde(NewPath));
  NewPath:= mbExpandFileName(Trim(FPathEdit.Text));
```
`NewPath` в первой строке пустой. Там должен быть `Trim(FPathEdit.Text)`, а уже во второй `NewPath`.

TheNicker: Fixed

PR #1352, opened 2023-11-16
ADD: custom directory separator when copying file names to clipboard

TheNicker:
Usage example:
When pressing `Ctrl+Shift+Z` under `Windows`, this will copy the file full path name using `slashes` and not `backslashes`. 
This is beneficial for working with cross platform development, CMake, etc.. 

image

### DIFF
--- a/src/umaincommands.pas
+++ b/src/umaincommands.pas
@@ -73,7 +73,7 @@ type
    procedure DoPanelsSplitterPerPos(SplitPos: Integer);
    procedure DoUpdateFileView(AFileView: TFileView; {%H-}UserData: Pointer);
    procedure DoCloseTab(Notebook: TFileViewNotebook; PageIndex: Integer);
-   procedure DoCopySelectedFileNamesToClipboard(FileView: TFileView; TypeOfCopy: TCopyFileNamesToClipboard);
+   procedure DoCopySelectedFileNamesToClipboard(FileView: TFileView; TypeOfCopy: TCopyFileNamesToClipboard; const Params: array of string);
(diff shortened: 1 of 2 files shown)


src/umaincommands.pas
```diff
@@ -666,8 +667,9 @@ procedure TMainCommands.DoCopySelectedFileNamesToClipboard(FileView: TFileView;
         case TypeOfCopy of
           cfntcPathAndFileNames, cfntcJustFileNames: FileNameToAdd:=SelectedFiles[I].Name;
         end;
-
-        sl.Add(PathToAdd + FileNameToAdd);
+        if (GetParamValue(Params, 'separator', separator)) then
+           sl.Add(ReplaceDirectorySeparator(PathToAdd + FileNameToAdd, separator[1]))
```
alexx2000, line 671: It will crash if user pass empty separator.

src/umaincommands.pas
```diff
@@ -666,8 +667,9 @@ procedure TMainCommands.DoCopySelectedFileNamesToClipboard(FileView: TFileView;
         case TypeOfCopy of
           cfntcPathAndFileNames, cfntcJustFileNames: FileNameToAdd:=SelectedFiles[I].Name;
         end;
-
-        sl.Add(PathToAdd + FileNameToAdd);
+        if (GetParamValue(Params, 'separator', separator)) then
+           sl.Add(ReplaceDirectorySeparator(PathToAdd + FileNameToAdd, separator[1])) 
+	    else sl.Add(PathToAdd + FileNameToAdd);
```
alexx2000, line 672: Don't use Tab symbol.

PR #1398, opened 2023-12-17
Update doublecmd.iss

VenusGirl:
Add Korean


alexx2000:
DC use Inno Setup 5 for compatibility with Windows XP and it does not have Korean translation. I don't know is IS 6 language file is compatible with IS 5 or not.

VenusGirl:
The Inno Setup 5 script encodes differently in Inno Setup 6 (UTF8 BOM).
That's why it's not compatible.

This is the Inno Setup 5 Korean script.
The encoding language code is ANSI (CP949).
[Korean.zip](https://github.com/doublecmd/doublecmd/files/13704985/Korean.zip)

PR #1432, opened 2024-01-06
FIX: macOS/darwin defaults RunTermCmd Terminal.app location

sriegler:
In recent major macOS versions the Terminal.app was moved out of the `/Application` directory into the SIP protected `/System/Application` directory.

Therefore the default installation config for macOS would result in a misconfiguration and the F9 shortcut for `cm_RunTerm` calls a missing target.

### DIFF
--- a/src/platform/uOSUtils.pas
+++ b/src/platform/uOSUtils.pas
@@ -57,7 +57,7 @@ const
   faFolder = S_IFDIR;
   ReversePathDelim = '\';
   {$IF DEFINED(DARWIN)}
-  RunTermCmd = '/Applications/Utilities/Terminal.app';  // default terminal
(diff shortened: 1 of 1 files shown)


rich2014:
@sriegler 

currently, DC has used `NSWorkspace.sharedWorkspace.fullPathForApplication()` in `uGlobs` to automatically obtain the path of Terminal.

for the first installation of DC, the user does not set the Terminal path in `Options-Tools-Terminal`, DC automatically sets its path.

if the user sets the Terminal path by himself, the user's setting is used.

if the user has set an old path before, the user can clear the path, and DC will automatically use the correct path.

alexx2000:
#1595

PR #1475, opened 2024-02-10
doublecmd.hu.po

dabzse:
HU language file updated: 2024-02-10


alexx2000:
This pull request has conflicts.

dabzse:
I'm so sorry about this late answer, I just doing some pushes almost every day with some code challenges and solvings. I did't see the notification.
I think the problem was the same branch name what I used earlier.
from now I will use: hu_<yyyymmdd>  to avoid any conflicts.
maybe I will do it today, but not sure. maybe tomorrow.

I'm not a professional git user, but will try to be better.

thank you!

PR #1509, opened 2024-03-07
Render Sync Dirs with custom font

lars18th:
The files listed in the Sync Dirs will be rendered using the font configured for the Search Results

### DIFF
--- a/src/fsyncdirsdlg.pas
+++ b/src/fsyncdirsdlg.pas
@@ -884,6 +884,7 @@ var
   s: string;
 begin
   if (FVisibleItems = nil) or (aRow >= FVisibleItems.Count) then Exit;
+  FontOptionsToFont(gFonts[dcfSearchResults], MainDrawGrid.Canvas.Font);
   with MainDrawGrid.Canvas do
   begin
     r := TFileSyncRec(FVisibleItems.Objects[aRow]);


lars18th:
This patch implements the enhancement #1508 .

Skif-off:
Forgive my curiosity, why the font for the Search Results? It seems Main Font looks more logical.

lars18th:
> why the font for the Search Results?

Because this Sync Dirs window is a new independent window, and it's similar to the Search Results window. That's the reason.

In any case, my point of view is that the font redering of doublecmd is ugly in some systems. And that's the reason because it has the option to configure some Fonts. But in the current implementation some texts aren't configurable (using then the default system font that could be ugly). So, because it's easy (in the source code) to change/select the redering of the font, I suggest to merge this patch. After, we can review if more fonts are using the default one and we can patch the code. And at the end, if we want then we could add more options in the configuration for specific areas. What do you think?

Skif-off:
Yes, I understand the reason for the proposed change, but the contents of the SyncDirs window are more similar to the contents of file panels and maybe the corresponding font appearance is more expected.
Maybe an additional option instead of a hard value would be more useful, I am not sure.

lars18th:
Hi @Skif-off ,

I understand your point of view. However, I feel that SyncDirs window are more similar to Search window because is a _NEW_ window. The Contents file panels are in the same Main window.

But, perhaps you can agree that this is a _secondary_ question. The main question is which Fonts can be default and not configurable. My sugestion is to follow this strategy:

1. Identify all non-configurable fonts, and add the code to use some alternative (aka configurable) font.
2. After all the fonts are covered, then review which one requires to be configured individually.

With this strategy will be more fast to develop it because:
- The code is easy to review because all configurable fonts are using the `FontOptionsToFont()` function.
- When all fonts are covered, then it will be more easy to group them in the configuration.

So, my suggestion is to help the design.
What you think?

alexx2000:
> I feel that SyncDirs window are more similar to Search window because is a _NEW_ window.

This is a controversial point. User change search results font but also SyncDirs font changed. Maybe it was a mistake to add search results font configuration and both search results and SyncDirs should use main file panels font.

lars18th:
> Maybe it was a mistake to add search results font configuration and both search results and SyncDirs should use main file panels font.

No please! In Windows, the current font differentiation is very valuable. Using a HDPI monitor it's preferable to use different fonts and different sizes. So, please NO, NO, NO.

If all of you think that this requires more effort to discuss about which fonts to group, them we can do it. However, IMHO a simple change of the Description in the Configuration Window will be sufficient: "Search Results/SyncDirs/...XX..." (and place other fronts to add to the list in XX). Which the problem to configure a "group of fonts"?

lars18th:
Hi @alexx2000 ,

Any comment about the best way to commit this change? The current "Main File Panel Font" in Windows with HDPI is not a good solution for the "Sync Dirs Font". Therefore, some solution is necessary. What you prefer?

lars18th:
Hi @alexx2000 ,

> > I feel that SyncDirs window are more similar to Search window because is a _NEW_ window.
> 
> This is a controversial point. User change search results font but also SyncDirs font changed.

This is not a controversial point if the Configuration of Fonts names them as: `Search Results / Sync Dirs Font`.

Please, note that the problem is this: In Windows (perhaps in other OS too) when using large HDPI modes, the "Main File Panel Font" needs to be different than the fonts in NEW Windows, like Search and Sync Dirs. Therefore, we _really_ need to configure the Sync Dirs font.

And IMHO the most simple solution is to _merge_ "Search Results" and "Sync Dirs" font configuration. But if you prefer to add another entry in the Fonts panel for the "Sync Dirs" then that's fine with me.

So could you please comment on your point of view?
Regards.

lars18th:
Hi @alexx2000 ,

We're still waiting for a _solution_ to this problem. For each new DC release I'm compiling a custom version with this patch. I want to reiterate: "**_With the current implementation the Sync Dirs font used is ugly on Windows_**". So please, add a comment.

Thank you.

lars18th:
Hi @alexx2000 ,

No comments about the problem of the **uggly** font in the Sync Dirs using Windows?

I suggest adding this _simple_ patch. And then, if you find it necessary to rewrite the font selection, then do it. But in the meantime, we (as users) with _every_ new release we must need to recompile with this PR because the font in the Sync Dirs window becomes uggly because of the default font used.

Please. 🙏

alexx2000:
Hi, I think this font can be renamed to `File lists` font and used in the search results, sync dirs and multi-rename tool.
But the patch is not quite correct. Font can be applied only one time on form create.

Skif-off:
@alexx2000

> I think this font can be renamed to File lists font and used in the search results, sync dirs and multi-rename tool.

This is not very convenient, because in this case we will not be able to specify the font styles separately.

P.S. A rhetorical question :)

> With the current implementation the Sync Dirs font used is ugly on Windows.

I don't quite understand it: there are many places in Windows where the default font from the system settings is used and why not replace it in the system settings?

alexx2000:
> This is not very convenient, because in this case we will not be able to specify the font styles separately.

Yes, that's why I don't sure. But create a separate font option for each dialog is too much. I have never seen anything like this in programs. TC doesn't have many font settings too.

Skif-off:
Maybe two instead three, sync dirs and search results & multi-rename tool?
Or delete some of them from the settings section, saving for manual change?

> I have never seen anything like this in programs.

This is rare, but it occurs, but unfortunately, I couldn't remember the names.

lars18th:
Hi @alexx2000 and @Skif-off ,

- Why DC needs the options to configure fonts? Because it's a Multiplatform tool.
- Why this specific window is uggly in Windows? Because at time using HDPI monitors you need to configure several aspects of the UI inside Double Commander.
- So, if you don't want to add support for _each_ dialog, then group them together: Sync Dirs and Multi-rename are of the same dialog type..

I hope this helps you decide on a good solution to the problem.

PR #1571, opened 2024-04-10
Update multiarc.ini

JonnyTech:
Additional extensions for `7z (ro)` taken from https://7-zip.org/index.html - removed `arj` - added `ima` and `img`

### DIFF
--- a/default/multiarc.ini
+++ b/default/multiarc.ini
@@ -23,7 +23,7 @@ Debug=0
 [7Z (ro)]
 Archiver=7z
 Description=7-Zip - www.7-zip.org
-Extension=cab,z,taz,lzh,lha,iso,wim,swm,dmg,xar,hfs,ntfs,fat,vhd,mbr
+Extension=apfs,ar,cab,chm,cpio,cramfs,dmg,ext,fat,gpt,hfs,ihex,ima,img,iso,lha,lzh,lzma,mbr,msi,nsis,ntfs,qcow2,rar,rpm,squashfs,swm,taz,udf,uefi,vdi,vhd,vhdx,vmdk,wim,xar,z
 Start=^-------------------
 End=^-------------------
 Format0=yyyy tt dd hh mm ss aaaaa zzzzzzzzzzzz pppppppppppp  n+


Skif-off:
You went too fast :)

chm - xCHM, for example.
cpio - cpio.wcx.
ext - ??? Extended File System? Well...
lzma - zip.wcx?
msi - Wine.
rar - unrar.wcx.
rpm - rpm.wcx.

alexx2000:
And why ARJ removed? I think most people will not install a standalone ARJ archiver. But can have 7Z.

JonnyTech:
> You went too fast :)
> 
> chm - xCHM, for example. cpio - cpio.wcx. ext - ??? Extended File System? Well... lzma - zip.wcx? msi - Wine. rar - unrar.wcx. rpm - rpm.wcx.

As stated in my OP, they were taken directly from the 7z home page as being supported by the software. I assume that if a plugin fails then it can faildown to try 7z.

JonnyTech:
> And why ARJ removed? I think most people will not install a standalone ARJ archiver. But can have 7Z.

Further down in `multiarc.ini` there is `arj` listed in another section. Shall I remove that entire section?

j2969719:
why remove any of them?  in most cases, the 7z addon will work out of the box, but if the user wants to use arj instead of 7z, then he can simply change the settings, and not search for the addon on the internet, besides, the addon for arj also works with ctrl+pgdn

Skif-off:
@JonnyTech 

> they were taken directly from the 7z home page

Yes, I understand, but I'm not sure that it makes sense to take what is already occupied by pre-installed plugins.
And CHM? MSI? These are ordinary files and should be opened in the associated program in the vast majority of cases.


@j2969719 

> afaik nsis is not an extension, but nullsoft installers (some exe files)

Sorry, I missed it, of course it should be separate and with a file signature(s) (I mean `ID` & `IDPos`). Or leave this decision to the users.


P.S. It's funny, CHM files were used in the [DC's help](https://doublecmd.github.io/doc/en/multiarc.html#openassociation) as an example.

PR #1710, opened 2024-06-12
Make DirHistoryCount configurable in misc settings.

kongomongo:
hi there,

as per https://github.com/doublecmd/doublecmd/discussions/1273 I tried looking at the source code and find out how to implement this rather simple change.

I have put the setting in the miscellaneous form and also in the settings xml under misc...

Please bear with me I have 0 experience with Lazarus. I can't for the life of me get the forms editor to append my label + spin edit after "default single-byte text encoding". Please help me here.

Also translations where auto-inserted, is that the way to go? Or should those be left out?

Please advise and I can amend the PR.

Thanks!


Skif-off:
Double Commander uses the same value and the same functions for any history, not only for directory history.
See `cMaxStringItems`.

P.S. I think there is no urgent need for an option in the settings section. (The list of parameters that can only be changed manually is described [here](https://doublecmd.github.io/doc/en/configxml.html).)

kongomongo:
> Double Commander uses the same value and the same functions for any history, not only for directory history. See `cMaxStringItems`.

I am not sure what you are saying. gDirHistoryCount is currently in main, it is not my invention. 

```
procedure SetDefaultNonConfigGlobs;
...
  gDirHistoryCount := 30;
```

cMaxStringItems however is a const, you cannot change it and it is not configurable.

> 
> P.S. I think there is no urgent need for an option in the settings section. (The list of parameters that can only be changed manually is described [here](https://doublecmd.github.io/doc/en/configxml.html).)

OK that would be fine with me, having no GUI and only set it via config. In which section should the gDirHistoryCount be situated? Then I can amend the PR.

Thanks!

Skif-off:
I thought you want to change the global value :) I did not know (or forgot) that the number of records of the history that is displayed after the "cm_DirHistory" command has a separate limiter.

> cMaxStringItems however is a const, you cannot change it and it is not configurable.

I guess that's the least of the problems...

> In which section should the gDirHistoryCount be situated?

`<History>` in `doublecmd.xml` and the "Configuration" settings section (in perspective).

I would probably try something like this
```diff
--- uglobs.pas
+++ uglobs.pas
@@ -305,7 +305,6 @@
   gToolbarPathModifierElements: tToolbarPathModifierElements;
 
   gRepeatPassword:Boolean;  // repeat password when packing files
-  gDirHistoryCount:Integer; // how many history we remember
   gShowSystemFiles:Boolean;
   gRunInTermStayOpenCmd: String;
   gRunInTermStayOpenParams: String;
@@ -529,6 +528,7 @@
   gSaveFileMaskHistory,
   gSaveVolumeSizeHistory,
   gSaveCreateDirectoriesHistory: Boolean;
+  gDirHistoryCount:Integer;
   gSortOrderOfConfigurationOptionsTree: TSortConfigurationOptions;
   gCollapseConfigurationOptionsTree: TConfigurationTreeState;
   
@@ -2003,6 +2003,7 @@
   gSaveFileMaskHistory := True;
   gSaveVolumeSizeHistory := True;
   gSaveCreateDirectoriesHistory := True;
+  gDirHistoryCount := 30;
   gPluginInAutoTweak := False;
   gWCXConfigViewMode := wcvmByPlugin;
 
@@ -2237,7 +2238,6 @@
   { - Not in config - }
   gHelpLang := '';
   gRepeatPassword := True;
-  gDirHistoryCount := 30;
   gFirstTextSearch := True;
   gErrorFile := gpCfgDir + ExtractOnlyFileName(Application.ExeName) + '.err';
   DefaultDateTimeFormat := FormatSettings.ShortDateFormat + ' hh:nn:ss';
@@ -2967,6 +2967,7 @@
     gSaveFileMaskHistory := GetAttr(Root, 'History/FileMaskHistory/Save', gSaveFileMaskHistory);
     gSaveVolumeSizeHistory := GetAttr(Root, 'History/VolumeSizeHistory/Save', gSaveVolumeSizeHistory);
     gSaveCreateDirectoriesHistory := GetAttr(Root, 'History/CreateDirectoriesHistory/Save', gSaveCreateDirectoriesHistory);
+    gDirHistoryCount := GetValue(Root, 'History/DirHistoryCount', gDirHistoryCount);
     gSortOrderOfConfigurationOptionsTree := TSortConfigurationOptions(GetAttr(Root, 'Configuration/SortOrder', Integer(scoAlphabeticalButLanguage)));
     gCollapseConfigurationOptionsTree := TConfigurationTreeState(GetAttr(Root, 'Configuration/TreeType', Integer(ctsFullExpand)));
 
@@ -3614,6 +3615,7 @@
     SetAttr(Root, 'History/FileMaskHistory/Save', gSaveFileMaskHistory);
     SetAttr(Root, 'History/VolumeSizeHistory/Save', gSaveVolumeSizeHistory);
     SetAttr(Root, 'History/CreateDirectoriesHistory/Save', gSaveCreateDirectoriesHistory);
+    SetValue(Root, 'History/DirHistoryCount', gDirHistoryCount);
     SetAttr(Root, 'Configuration/SortOrder', Integer(gSortOrderOfConfigurationOptionsTree));
     SetAttr(Root, 'Configuration/TreeType', Integer(gCollapseConfigurationOptionsTree));
 
```
(I didn't check it, I just wrote it on my phone and created a patch.)

As I understand, DC remembers up to 255 directories visited, but when closing DC, only 50 (`cMaxStringItems`) are preserved.

I think the best way is to ask @alexx2000 to add the ability to control the size of the history (in addition to the maximum size limit, there are some other details). But I don't have a concrete idea yet.

kongomongo:
OK i made a "no ui" version of this PR https://github.com/doublecmd/doublecmd/pull/1736

I'd close this as soon as @alexx2000 says how to proceed.

Thanks in advance!

PR #1736, opened 2024-06-29
Make DirHistoryCount configurable (via config file only).

kongomongo:
as per https://github.com/doublecmd/doublecmd/pull/1710 this is a new PR which simply makes the DirHistoryCount changable via config file.

That means while double commander is running you can now have as many history items as you like in memory. (default is currently 30)

When you close and reopen double command there is still a limit of how many items get saved and reloaded via cMaxStringItems (not configurable currently, default is 50 currently).

Would be great if we can get this merged ;-)


kongomongo:
pinging @alexx2000 

Thanks in advance!

Skif-off:
It seems that you read the value twice and save it twice?

kongomongo:
> It seems that you read the value twice and save it twice?

you are of course correct. sorry for the confusion. Fixed.

kongomongo:
any chance to get this into master?

alexx2000:
I think it should be stored like:

`SetAttr(Root, 'History/DirHistory/Count', gDirHistoryCount);`

`GetAttr(Root, 'History/DirHistory/Count', gDirHistoryCount);`

kongomongo:
thank you!

lorenzo-gi:
I really appreciate this improvement. Allowing also cMaxStringItems to be changed would be also great

kongomongo:
> I really appreciate this improvement. Allowing also cMaxStringItems to be changed would be also great

I agree, I tried another PR here: https://github.com/doublecmd/doublecmd/pull/1867

PR #1758, opened 2024-07-20
UPDATE: Hungarian translation

dabzse:
hopefully I didn't miss anything and got it right the first time with this branch.
I'm almost sure there are some typos / small misstranslations, but they will be corrected as well.

doublecmd.pot isn't updated on the first lines:
"Project-Id-Version: Double Commander 1.1.0 alpha\n"
"POT-Creation-Date: 2022-11-15 11:15+0300\n"

about the screnshots:

dc_1
BackGround: > possible should be Background (without :), but didn't check it yet
dc_1

dc_2
I have translated as bg1 and bg2 so far; for me is much cleaner to number the first as well if there are more
dc_2


alexx2000:
I applied a formatted version. Lazarus po-file formatting differs from PoEdit formatting. So after editing in the PoEdit, file has many useless changes (different formatting of a long lines).

PR #1810, opened 2024-09-03
Some additional functions for Lua

Skif-off:
DC.ExpandVar, DC.GetPluginField, DC.GoToFile, DC.ExpandEnv;
SysUtils.FileSetTime, SysUtils.GetFileProperty, SysUtils.RemoveDirectory.

### DIFF
--- a/src/uluapas.pas
+++ b/src/uluapas.pas
@@ -39,7 +39,10 @@ uses
   Forms, Dialogs, Clipbrd, LazUTF8, LCLVersion, uLng, DCOSUtils,
   DCConvertEncoding, fMain, uFormCommands, uOSUtils, uGlobs, uLog,
   uClipboard, uShowMsg, uLuaStd, uFindEx, uConvEncoding, uFileProcs,
-  uFilePanelSelect, uMasks, LazFileUtils, Character, UnicodeData;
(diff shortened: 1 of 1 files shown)


alexx2000:
`TFileSystemFileSource.CreateFileFromFile` should be in the `try..except`, it genarates exception when file not found.

Skif-off:
Thanks! Please take a look now: I tried to fix it and also changed the algorithm of these two functions a bit.

alexx2000:
Algorithm is better now. But I don't see a fix. It should be like:
```
try
  AFile := TFileSystemFileSource.CreateFileFromFile(lua_tostring(L, 1));
except
  lua_pushnil(L);
  Exit;
end;
```

Skif-off:
It seems to be fixed now.

Perhaps this
```pascal
   if AIndex > 9 then
    begin
      lua_pushnil(L);
      Exit;
    end;
    try
      ...
```
should be replaced with something like
```pascal
    if AIndex < 10 then
    begin
      try
        ...
```

Skif-off:
Done.

P.S. I also thought about getting the total number of items and the number of selected ones, but I'm not sure.

PR #1835, opened 2024-09-13
UPD: RAR progressbar

j2969719:


j2969719:
btw, do totalcmd plugins update progressbars correctly in this case? docs seems to says the opposite:
>Set size to negative percent value (-1..-100) to directly set first percent bar, -1000..-1100 for second percent bar (-1000=0%).

https://ghisler.github.io/WCX-SDK/tprocessdataproc.htm

j2969719:
invalid

PR #1838, opened 2024-09-15
ADD: MultiArc - AskHistory support

j2969719:


src/fpackdlg.pas
```diff
@@ -398,6 +411,8 @@ procedure TfrmPackDlg.SwitchOptions(ArcTypeChange: Boolean); // Ugly but working
           else
             sCmd:= FAdd;
 
+          lblPacker.Caption:='MultiArc [' + FPacker + ']';
```
alexx2000, line 414: I don't like the `MultiArc` prefix, I think it will be better without it. We can distinguish what is used by square brackets and  `.wcx` extension.

j2969719: ok, i just added this label because of the issue when one extension is used several times, to see which packer will actually be used. 

btw, maybe instead of for loops, extensions should be stashed to some stringlist with userdata or something, to remove duplicates since dc doesnt use them anyway

j2969719:
also, it seems i didnt add it to ComputeSignature

alexx2000:
> also, it seems i didnt add it to ComputeSignature

It is not a configuration so I think `ComputeSignature` shouldn't take it into account.

j2969719:
ok, removed

alexx2000:
Still `lblPacker` looks visually bad. I don't know what to do. Maybe is better to show this info in the `Hint` for `btnConfig` or some other control.

j2969719:
dunno btnconfig is now disabled if the packer does not know how to configure. im cool with removing the label, especially since it has little to do with prams history

though imho something needs to be done with those ext duplicates

PR #1848, opened 2024-09-21
UPD: Variables in file names for all internal commands

Skif-off:
`cm_SaveTabs`, `cm_LoadTabs`, `cm_LoadList`, `cm_ChangeDir` (legacy).

### DIFF
--- a/src/umaincommands.pas
+++ b/src/umaincommands.pas
@@ -1934,15 +1934,18 @@ begin
   //    Let's assume that if we have an "=" sign, it's can be a legacy usage but one with actual parameters.
   if (length(Params)>0) then
   begin
-    sOutputTabsFilename := GetDefaultParam(Params);
-    if pos('=',sOutputTabsFilename)<>0 then
+    sValue := GetDefaultParam(Params);
+    if pos('=',sValue)<>0 then
     begin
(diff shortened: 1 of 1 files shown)


Skif-off:
@alexx2000
What do you think about these changes?

alexx2000:
I'm not sure yet that it is a good idea to use it everywhere. `RemoveQuotation` still can broke some file names, for example `hello\ people`. Maybe it is better to use here a simple function instead, which removes only a single or double quotes around file name.

Skif-off:
> I'm not sure yet that it is a good idea to use it everywhere.

I'm not sure too.
Currently, `cm_SaveTabs`, `cm_LoadTabs` and `cm_LoadList` cannot work with DC/environment variables (only full file names), so I just wanted the variables to work the same with all commands.

Skif-off:
> Maybe it is better to use here a simple function instead, which removes only a single or double quotes around file name.

Maybe, but what about backslash `\` in Linux? `PrepareParameter` > `ReplaceVarParams` > `QuoteStr` > `EscapeNoQuotes` > `EscapeString`.
Should we always use `%"0`?

alexx2000:
> Should we always use %"0?

Yes (for internal commands).

PR #1860, opened 2024-09-24
ADD: MultiArc - Show as normal files

j2969719:


Skif-off:
I am not sure that it's really needed, because we already have a simple way...
But in any case, as I see, you forgot about [src/platform/upixmapmanager.pas#L1815](https://github.com/doublecmd/doublecmd/blob/1bfec1424ca7c3269ed53b434749513af0574603/src/platform/upixmapmanager.pas#L1815).

j2969719:
tnx
some files (diff/patch for patchfs, hash for 7zz etc) are easier to register by extension, but then you will have to open them not as an archive only from the context menu

+atm multiarc does not open addons registered only by extension by ctrl+pgdn

Skif-off:
> multiarc does not open addons registered only by extension by ctrl+pgdn

I don't remember :)) And I can't check.
It seems the file will open anyway, I found this way: `cm_OpenArchive` > `ChooseArchive(frmMain.ActiveFrame, FileSource, aFile, True);` > `GetArchiveFileSource(aFileSource, aFile, EmptyStr, bForce, False);` > `GetArchiveFileSourceDirect(SourceFileSource, ArchiveFile.FullPath, ArchiveType, ArchiveSign, IncludeHidden);` with [src/filesources/uarchivefilesourceutil.pas#L68](https://github.com/doublecmd/doublecmd/blob/0261d283f4c0b6f2252cf541e39895ab09b73b17/src/filesources/uarchivefilesourceutil.pas#L68). But I guess I misunderstood something.

PR #1863, opened 2024-09-24
UPD: MultiArc - If day or month is not set then make mtime invalid

j2969719:


j2969719:
when using addon without date (eg innosetup addon from neighboring pullrequest) dynamic parser set date to 2000-01-01

PR #1878, opened 2024-09-30
UPD: Always put files with empty field at the of filelist when sortin…

j2969719:
…g by plugins

### DIFF
--- a/src/ufilesorting.pas
+++ b/src/ufilesorting.pas
@@ -551,7 +551,12 @@ end;
 
 function ICompareByVariant(Value1, Value2: Variant; bSortNegative: Boolean):Integer;
 begin
-  if VarIsType(Value1, varString) then
+  // Put clear values last.
+  if VarIsClear(Value1) then
+    Exit(1)
+  else if VarIsClear(Value2) then
+    Exit(-1)
+  else if VarIsType(Value1, varString) then
     Result := CompareStrings(Value1, Value2, gSortNatural, gSortSpecial, gSortCaseSensitivity)
   else if Value1 = Value2 then
     Exit(0)


Skif-off:
Will the sorting direction be taken into account or not?

alexx2000:
In this patch no.

Skif-off:
It doesn't seem very convenient.

j2969719:
my pov: when sort by plugin value personnaly iam only interested in files with result. currently when comparing strings, files with empty values ​​always jump to the beginning/end of the list and if the list is big you need to spend more time scrolling to the appropriate files

Skif-off:
But an empty value is also a result.

Skif-off:
Hmm. I tried using it and I'm not sure it's better. (And the optional parameter will probably "eat" up a lot of time...)

j2969719:
are there issues in some usecase or is it just a matter of habit?

Skif-off:
When I change the sort direction, I expect a certain result at the top or bottom of the list.
Depending on the situation, sometimes it is more convenient to have files with an empty line at the top of the list, sometimes vice versa, because an empty line is not always an error/problem.
For example, I use [symlinkwdx.lua](https://github.com/j2969719/doublecmd-plugins/blob/master/plugins/wdx/scripts/symlinkwdx.lua) with the "Target" field quite often to distinguish between regular files and links: sometimes I need one, sometimes I compare the other.
Is all this "usecase" or "habit"? :)

P.S. The difference from sorting by extension, if there are files without an extension, is a bit unexpected.

j2969719:
empty != incorrect, nil is returned in line 52, which is equivalent to the error code ft_fileerror/ft_fieldempty in c plugins. try returning "", but welp i can also add option in the config if this realy necessary

Skif-off:
I made a mistake, I used `GETFILELINKTO` in this copy. (There's a [weird problem](https://doublecmd.h1n.ru/viewtopic.php?p=44879#p44879) with `symlinkwdx.lua`.)

> i can also add option in the config if this realy necessary

Won't that slow it down too much?

j2969719:
i think its just a weird issue with comparing a string to null, this shouldnt happen with this patch

its just an extra boolean comparison in the condition, but ideally id like to go without it if its not necessary

also thought maybe for incorrect files in the list it should show instead of an empty string eg ?? so that they could be distinguished, but in reality thats a bit annoying with regular font/color

Skif-off:
> its just an extra boolean comparison in the condition, but ideally id like to go without it if its not necessary

I don't know, I don't really understand the implications in the long run, so I opened my mouth :))

> also thought maybe for incorrect files in the list it should show instead of an empty string eg ?? so that they could be distinguished

Not very good idea, nothing is nothing. I think this is the plugin's job.

j2969719:
dont quite get what you mean by the plugin job because it did its job when it returned the result. atm w/o clear indication nothing is both an empty string and an incorrect file. while with patch, eg when sorting by album with audioinfo.wdx, audio files with an empty tag will be sorted, and covers, lyrics, playlists or just some other random crap will always be at the bottom. from my pov, its more convenient this way, but dunno w/o clear indication, questions may arise about why exacly this happens

can you check how other commanders behave on win when you have time? tried adding audioinfo.wdx on wine, but it doesnt show up in the plugin list for some reason when i create a columnset

Skif-off:
> dont quite get what you mean by the plugin job because it did its job when it returned the result.

I mean that the program shouldn't add "??" or anything like that.

> atm w/o clear indication nothing is both an empty string and an incorrect file. while with patch, eg when sorting by album with audioinfo.wdx, audio files with an empty tag will be sorted, and covers, lyrics, playlists or just some other random crap will always be at the bottom. from my pov, its more convenient this way

But if you are looking for files with an empty tag, it is less convenient :)
I'm trying to say that an empty field is also useful information and it's certainly illogical to push it into the far side in any case.

> can you check how other commanders behave on win when you have time?

TC and DC behave the same way.

> tried adding audioinfo.wdx on wine, but it doesnt show up in the plugin list for some reason when i create a columnset

Check the list of dependencies, I saw `ole32.dll` - OLE/ActiveX, but I don't know why FP/Lazarus use ActiveX in this case(s). Most likely it won't work as is, out of the box.

j2969719:
>But if you are looking for files with an empty tag, it is less convenient :)
I'm trying to say that an empty field is also useful information and it's certainly illogical to push it into the far side in any case.

erm, you mean when you search for files with an empty tag to fill them its more convenient to have the incorrect files sorted in one pile? again, with the patch files with an empty tag are sorted as strings.

Skif-off:
> with the patch files with an empty tag are sorted as strings.

If I use `if sTag == "" then return nil end` then I will have problem?

Maybe I'm a little stupid or maybe I don't fully understand how it should work or maybe I don't understand how convenience in some particular case would be convenient in all cases :))

P.S. I forgot to ask,
```pascal
else if VarIsType(Value1, varString) and not VarIsClear(Value2) then
```
why is the second part of the condition added? `VarIsClear(Value2)` is already in the previous condition.

j2969719:
it will be more practical to just test it more and provide feedback on how and when it is inconvenient in your workflow tbh im starting to get the impression that im just increasing the comment counter by repeating the same thing:
files will be placed at the end of the list when the variant value was not set i.e. if ContentGetValue returned an error code (or nil in the case of lua), if, as with tags for example, the plugin returned value as an empty string, it will be sorted as a string. 

additional condition added since the patch is now optional and disabled by default, so that the clean variant doesnt give undefined results(when sort presed several times) when passed to CompareString, that you mentioned on the forum

Skif-off:
> it will be more practical to just test it more

Yesterday (again). DC alpha, DC from your branche, TC + audioinfo & TCMediaInfo, poppler_info & xPDFSearch (1.11); folders with homogeneous content and not.

PR #1880, opened 2024-10-01
ADD: MultriArc - comment parsing

j2969719:


j2969719:
ffff

alexx2000:
Total Commander [does not support](https://ghisler.github.io/WCX-SDK/theaderdataex.htm) the `Cmt*` fields at this moment. Do you plan to use these fields in your plugins?

j2969719:
maybe only for zpaq, but it would probably be easier to be consistent with tc in this case
should this property be added only to multiarc filesource? similar thing with symlinks, multiarc can parse `->` and think they can be displayed if thereare any

alexx2000:
Technically I can enable `Cmt*` for plugins. But I don't know how necessary it is.

> should this property be added only to multiarc filesource?

Yes, if we don't plan to enable it for plugins.

j2969719:
sounds tempting, and bit7z lib seems to be able to also give comments for items, but dont think its worth wasting time just for the sake of my shitcode, it may also cause some wierd issues with existing tc plugins.. fparchive thing can parse zip comments?

on another note, iirc multiarc.wcx can also parse `c`, now im wondering what it does with it

j2969719:
>multiarc.wcx can also parse c, now im wondering what it does with it

>grep "'c'" ArcItem.cpp

mCD3c0e

Issue #1669, opened 2024-05-22
Proper 7z defaults for macOS


lprot:
### Version

Double Commander
Version: 1.2.0 alpha
Revision: 11332
Commit: afc3a9fc9
Build date: 2024/05/20
Lazarus: 3.99.0.0
FPC: 3.2.2
Platform: x86_64-Darwin-cocoa
OS version: Mac OS X 12.7.5

### Description

@alexx2000 on macOS, default archiver has to be not 7z but **7zz** with default packer extensions 7z,bzip2,gzip,tbz2,tpz,xz,wim:
![image](https://github.com/doublecmd/doublecmd/assets/2608545/2e40c935-aad1-4487-a0d8-576b39a03328)

Unpacker 7zz with extensions: apfs,ar,arj,cab,chm,cramfs,dmg,esd,ext,fat,gdp,gpt,hfs,ihex,ima,img,iso,lha,lzh,mbr,msi,nsis,ntfs,qcow2,split,squashfs,swm,taz,tzst,udf,uefi,vdi,vhd,vhdx,vmdk,xar,z,zst
![image](https://github.com/doublecmd/doublecmd/assets/2608545/a7d3ef10-cd14-4b25-a7b5-a4202c301bca)

This pull request can be cancelled https://github.com/doublecmd/doublecmd/pull/1571

Reference: CPP/7zip/Bundles/Format7zF/resource.rc

j2969719:
again, the supported format != extension. eg things like nsis can only be registered in multiarc using hex magic. assigning a multiarc packer to formats that support standard wcx plugins also makes little sense, since using an external archiver you can add new files to an existing archive at the root only.

lprot:
Here are the extensions that are valid in case we use 7zz. I do not see any problems with nsis association with 7zz. On maxOS there is no WCX supporting nsis.

j2969719:
again, nsis is not an extension, it is a nullsoft installer - a regular exe file. please make sure once again that the proposal is correct(eg i some have slight doubts about that funny split extension)

j2969719:
erm, i rather just mean dont add some random invalid extensions out of the box or duplicate zip.wcx associations

lprot:
which zip.wcx duplicate do you see here?

j2969719:
bzip2,gzip,tbz2

lprot:
where do you see those here?
[...]

PR #1940, opened 2024-10-24
ADD: MultiArc - update settings when defaults/platform changes

j2969719:


ags1234:
@j2969719 , what is this MultiArc that you keep updating files for?

j2969719:
https://doublecmd.github.io/doc/en/multiarc.html

Skif-off:
Small note: the help files mention the ability to import/export settings, but without much detail (I don't know what could be added there).

@j2969719  are you sure that forcing the `AutoConfigure` function is a good idea? If user is in the process of configuring several archivers, this can be inconvenient.

Perhaps it would be better to make changes in the `multiarc.ini` file as a separate PR?

j2969719:
autoconfig(search for an executable and disabling an addon if the executable is not found) will be forced when, for example, the config reports that it is saved under win, and the dc version is for linux, or when something was added/changed in default/multiarc.ini (in this case, existing default sections will be overwritten and backup copy will also be created). 

people are already starting to ask why im so obsessed with this multiarc thing and updating the default config with this patch implies a version bump in umultiarc.pas anyway

Skif-off:
I meant that `AutoConfigure` will process ALL archivers, not just the one being added in this moment :)

j2969719:
welp thats kind of the point of it. i mean to enable those that are currently available and disable those that are not, since user could have installed/removed some archivers since the last autoconf, but didnt touch this settings in dc, or doesnt know anything about them at all. and for those who tinker with these settings there will be a backup copy

ags1234:
So this is all about the config stored in the file `\settings\multiarc.ini`?

Skif-off:
@j2969719, this is a global function, i.e. it will enable what I have disabled and disable what I have not finished with yet :)

@ags1234, yes, about configure external archivers.

j2969719:
erm, do you assign executables at the very end?o_O version bump for multiarc will most likely occur even less often than for shortcuts.scf anyway

j2969719:
can someone clone and look at the ace addon? github diff seems to show two spaces before n+, but on my end this line seems to be the same length as the current one o_O

Skif-off:
> erm

This function does more than it is required to do in this case.

> github diff seems to show two spaces before n+

Maybe there is a problem with fonts or rendering? I tried looking in GitExtensions, the line lengths are the same there.

j2969719:
after #1669 i think on the contrary that autoconfiguration is too little in this life because the user can install the archiver whenever he wants and ideally it should just work. while tinkers should bear their burden with dignity [and pain]

PR #1960, opened 2024-11-05
ADD: Set archive time to newest file time

j2969719:


j2969719:
plot twist: packers that do not return directories will have the time unchanged because filesource for such directories sets the archive time

PR #1996, opened 2024-12-03
UPD: use -sfx switch for cli rar

j2969719:


Skif-off:
`*.run` is usually text file, isn't it?

j2969719:
¿
when sfx checkbox is enabled, dc sets the archive extension to `exe` on windows and to `run` on other platforms.
btw rar for linux with this switch for some reason creates archive with the extension `sfx`, regardless of the specified extension

Skif-off:
It was a question for "general development" :)

P.S. Yes, I know about `GetSfxExt`, but I've never tried packing with this plugin, so I didn't know about the `sfx` extension in some (all?) cases.

j2969719:
ithink its just rar's quirk, 7z doesnt have this issue

alexx2000:
Why only console RAR? WinRAR supports this option too.

j2969719:
iirc when testing on wine, winrar created sfx even without this switch. so i thought that the absence of this switch was intended since winrar is used there by default

PR #2016, opened 2024-12-19
ADD: Built-in editor menu item "Edit - Time/Date".

R0bur:
The implemented feature inserts current time and date into the current position of the edited text, like it Notepad in Windows does.

### DIFF
--- a/src/feditor.lfm
+++ b/src/feditor.lfm
@@ -820,6 +820,9 @@ object frmEditor: TfrmEditor
       object miGotoLine: TMenuItem
         Action = actEditGotoLine
       end
+      object miTimeDate: TMenuItem
+        Action = actEditTimeDate
+      end
       object N5: TMenuItem
         Caption = '-'
       end
@@ -1030,6 +1033,11 @@ object frmEditor: TfrmEditor
       Category = 'Edit'
       Caption = 'Goto Line...'
       ImageIndex = 16
+      OnExecute = actExecute
+    end
+    object actEditTimeDate: TAction
+      Category = 'Edit'
+      Caption = 'Time/Date'
       OnExecute = actExecute
     end
     object actEditFindPrevious: TAction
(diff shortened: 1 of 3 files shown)


Skif-off:
Is a hard coded template a good solution?

R0bur:
The 'ddddd' template will be replaced with a short date representation as it is set in the system locale settings. It should be exactly the same as in the Notepad implemented. But I tested it in Windows 10 only.

Skif-off:
Yes, I know how these characters will work. It has always seemed to me that using `yyyy.mm.dd hh:mm` or `yyyy.mm.dd hh:mm:ss` is more practical.

> It should be exactly the same as in the Notepad implemented.

I am not ready to consider this program authoritative :)
For example, Geany allows to [set a template](https://www.geany.org/manual/current/index.html#template-preferences) (for the "Insert date" command too). Or maybe without GUI like [DateTimeFormat](https://doublecmd.github.io/doc/en/configxml.html#DateTimeFormat).

Maybe it makes sense to rename "Time/Date" to "Insert date"?

P.S. In any case, I'm ready to wait for Lua support in the editor :)

R0bur:
> 
> > It should be exactly the same as in the Notepad implemented.
> 
> I am not ready to consider this program authoritative :) For example, Geany allows to [set a template](https://www.geany.org/manual/current/index.html#template-preferences) (for the "Insert date" command too). Or maybe without GUI like [DateTimeFormat](https://doublecmd.github.io/doc/en/configxml.html#DateTimeFormat).

[AkelPad](https://akelpad.sourceforge.net/) uses the same simple implementation as Notepad.

> Maybe it makes sense to rename "Time/Date" to "Insert date"?

All details are easier to fix when the basic feature will be implemented.

j2969719:
imo this is the most common use case and customizable template imo overengineering because even if you suddenly need another form, it is much faster to edit the date in the editor than to dig into the settings and study the docs on templates

and for more flexibility or some highly specialized use cases, perhaps lua may help in the future

Skif-off:
@R0bur

> AkelPad uses the same simple implementation as Notepad.

If you ask me "_Skif-off, who do you love more, hmm... Monica Bellucci or AkelPad?_", I'll be a little confused :))
But, in any case, see `DateInsertFormat` in `AkelFiles/Docs/AkelHelp-Eng.htm`.

@j2969719

> imo this is the most common use case and customizable template imo overengineering

You're probably right, I don't know.

PR #1862, opened 2024-09-24
Updated Slovak language

dodog:
Updated Slovak language file

PR #2092, opened 2025-01-18
Updated Slovak language

dodog:
New strings translated and fixed


alexx2000:
https://github.com/doublecmd/doublecmd/pull/1862#issuecomment-2375038540

dodog:
Yes, I have changed the settings in Poedit as you recommended last time. I even double-checked before making a PR. Is it possible, it is not applied to the file?
screenshot-2025-01-19_22-20-09

alexx2000:
Strange... As I understand it is a global setting. I don't know why it had not applied.

PR #2255, opened 2025-04-05
UPD: better thread-safe TFileSource reference counting

xBZZZZ:
1. works with latest freepascal ( https://gitlab.com/freepascal.org/fpc/source/-/commit/c3f80014b41db2cb4a6fa50a50c2548a03000124#note_2431141761 )
2. impossible to find IFileSource during constructing or destructing

note: there is code like: find {subclass of `IFileSource`} or create it if not found

https://github.com/doublecmd/doublecmd/blob/b7b47b21c17e47b35f60c972bdf242139b835a06/src/filesources/uarchivefilesourceutil.pas#L52-L61

that is not thread safe, multiple of {subclass of `IFileSource`} can be created if code like that runs on multiple threads


xBZZZZ:
todo: test plugins

PR #2294, opened 2025-04-30
FIX: LANG: zh_CN: Fix translations

rumia-san:
I found some wrong translation:
**婪**删除  -> should be **要**删除 
"个文件" -> should be “文件”

screenshot:
Screenshot 2025-05-01 010132
Screenshot 2025-05-01 010220

Could you help to check it? Thanks!


rich2014:
@rumia-san 

thx

Issue #2131, opened 2025-01-30
Random data corruption while copying files to SMB network share


codetheron:
### Version

Double Commander
Version: 1.1.22 gamma
Revision: 662
Commit: 6baafba86
Build date: 2025/01/19
Lazarus: 3.8.0.0
FPC: 3.2.2
Platform: aarch64-Darwin-cocoa
OS version: Mac OS X 15.2.0

Synology DS218+
DSM 7.2.2 update 2 (current version)

![Image](https://github.com/user-attachments/assets/7d7b3b93-bb44-42db-b5a0-66ad057672fa)

![Image](https://github.com/user-attachments/assets/93690caa-14db-46a7-ae57-449d8422cac1)

### Description

I'copied a big 7z archive of archived data from some old computer, splitted into several 2GB volumes onto my Synology NAS SMB share. As - when working with Window's Total Commander I tend to copy such data with verify option - and Double Commander has no 'verify option' I know of - I've decided to make checksums of those files - just in case.

To my surprise - some of the copied files have had bad checksums. I thought that I did something in wrong way, did it once again, and the results were similar.

Since then:
- I've copied before-mentioned files several times via Mac Finder, muCommander and  Marta and in all cases checksums were correct
- I've played with settings of my SMB share on my Synology and it seems to be set correctly - it's most recent DSM version (7.2.2) with SMB fixes for MacOS
- As I'm using Brew on my Mac - I've confirmed exact the same behavior on DC installed from official site

What's interesting - comparing by contents shows, that corrupted archives have zeroes starting from some address that looks as round in hex notation (as shown on attached screenshot).

What's more interesting - files are corrupted only by copying them to NAS, copying them from NAS to Mac is OK.

alexx2000:
#2018

codetheron:
PS @alexx2000 - on Mac OS there is no 'verify file' option in file copy dialog? Is there a way to enable it?

alexx2000:
Version 1.2.0 has this option (experimental).

codetheron:
Thanks, I will test this version against this bug too. SMB is recommended to use by Apple lately, so it would be great to get rid of this bug :)

codetheron:
This problem persists in 1.2.0 alpha (Revision: 11835, Build date: 2025/01/30). But it seems that copy with verify works correctly, but it's very slow:

Double Commander without verify:
speed: 112 MB/s
time: 12 min

Double Commander with verify:
speed: 34 MB/s
[...]

PR #2404, opened 2025-07-14
issues/2131

rseleven:
Possible fix for this issue [issues/2131](https://github.com/doublecmd/doublecmd/issues/2131). Needs testing. Haven't gotten to building and testing it myself yet.

In any case, verifying that both source and destination have APFS filesystems would be useful before using copyfile().


rich2014:
@rseleven 

copyfile() was only implemented in July this year, so #2131 is not caused by copyfile().

in addition, for file systems that do not support clone, it will automatically degenerate to the original copy function of DC.

PR #2747, opened 2026-02-13
FIX: Unicode normalization in Synchronize Dirs on MacOS (fixes #1752)

### LINKED ISSUE #1752: Compare function under Synchronize Dirs doesn't work with special characters (macOS x86)
### Version

Double Commander
Version: 1.1.16 gamma
Revision: 520
Commit: ca77fea98
Build date: 2024/06/22
Lazarus: 3.99.0.0
FPC: 3.2.2
Platform: x86_64-Darwin-cocoa
OS version: Mac OS X 12.7.5

### Description

When synchronizing directories on AFPS-formatted internal SSD to exFAT-formatted external USB drive, Compare does not work for filenames containing special (non-ASCII) characters like ä, ö, or é. Identical filenames with these characters are not correctly recognized by by Double Commander - Synchronize Dirs -> Compare function. When identical filenames contain only ASCII characters, they are recognized correctly by Double Commander - Synchronize Dirs -> Compare function. This has been non-functional for many versions.

ilabutin:
Fixes issue #1752 with incorrect behavior of Synchronize Dirs dialog when "special" symbols are used (like ä, ö, or é).

The DCOSUtils already have NormalizeFileName function that correctly converts Unicode strings from NFD to NFC. So, using it to have normalized name to be displayed in the list.

### DIFF
--- a/src/fsyncdirsdlg.pas
+++ b/src/fsyncdirsdlg.pas
@@ -239,7 +239,7 @@ uses
   uFileSystemFileSource, uFileSourceOperationOptions, DCDateTimeUtils, SyncObjs,
   uDCUtils, uFileSourceUtil, uFileSourceOperationTypes, uShowForm, uAdministrator,
   uOSUtils, uLng, uMasks, Math, uClipboard, IntegerList, fMaskInputDlg, uSearchTemplate,
-  LCLVersion, SysConst, DCStrUtils, uTypes, uFileSystemDeleteOperation, uFindFiles;
+  LCLVersion, SysConst, DCStrUtils, DCOSUtils, uTypes, uFileSystemDeleteOperation, uFindFiles;
 
 {$R *.lfm}
 
@@ -1390,6 +1390,7 @@ var
       i, j: Integer;
       f: TFile;
       r: TFileSyncRec;
+      fn: String;
     begin
       if sideLeft then
         fs := FFileSourceL.GetFiles(BaseDirL + dir)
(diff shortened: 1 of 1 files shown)


Skif-off:
It seems to me that similar feature must be optional: filenames _**look**_ the same, but they are _**not**_ the same and users need to understand what they see and what they do.

P.S. Search and correction: [filenameunwdx.lua](https://github.com/j2969719/doublecmd-plugins/blob/master/plugins/wdx/scripts/filenameunwdx.lua), the LuaJIT library must be installed and added in the [Plugins](https://doublecmd.github.io/doc/en/configuration.html#ConfigPlugins) settings section (DC distributives for Windows already contain the Lua library from the LuaJIT project).

alexx2000:
Under macOS, such names are considered the same. `NormalizeFileName` function has effect only under macOS so this change will not affect other systems. I was planning to make similar changes.

Skif-off:
What happens if there are two files in the folder with the letter "é", but U+00E9 in the first and U+0065 + U+0301 in the second?

alexx2000:
I guess on the native macOS file system such files cannot co-exist. While exFAT technically can have such files. But you have to make a special effort to create them. I hope this is a rare case.

Skif-off:
If it only affects macOS, then I don't see any problems for myself, but I checked the "U+00E9 vs. U+0065 + U+0301" case: Xubuntu/ext4 & Win7/NTFS see such files as different files.

ilabutin:
> I checked the "U+00E9 vs. U+0065 + U+0301"

I tried to create two files with these names on MacOS - and no luck. The following commands touch the same file:

    touch $'try\u00E9.txt'
    touch $'try\u0065\u0301.txt'

Skif-off:
I assume that in both cases, the `touch` utility called the same function, which converts the combined characters into the traditional macOS form.

I don't have any Apple devices, so I just created `tést.txt` and used the multi-rename tool & `filenameunwdx.lua` with a copy of it. (For control, just in case, I copied both filenames to clipboard, save them to a text file (UTF-16 BE), and checked them in a hex editor.)

PR #2760, opened 2026-02-20
FIX: main window loses focus after confirming file/folder deletion (fixes #2160)

### LINKED ISSUE #2160: After confirming a file/folder deletion the Double Commander window loses focus
### Version

Double Commander
Version: 1.1.22 gamma
Revision: 662
Commit: 6baafba86
Build date: 2025/01/19
Lazarus: 2.2.6.0
FPC: 3.2.2
Platform: x86_64-Linux-qt5
OS version: Debian trixie/sid

Widgetset library: Qt 5.15.15

### Description

Basically, the problem is in the description. Once confirming the deletion using the dialog the main window loses focus, and pressing keyboard shortcuts further actually happens in a totally different app, for example a browser, which is very annoying.

Image

arsinclair:
## Problem

After confirming a file or folder deletion in the confirmation dialog (clicking OK), the Double Commander window loses focus instead of keeping focus on the main window, requiring an extra click to continue.

## Root cause

1. **Dialog owner**: The delete confirmation dialog was created with `Application` as owner instead of the main form. On some platforms/widget sets (particularly Linux), this can prevent the LCL from correctly returning focus when the modal closes.
2. **No explicit focus restoration**: There was no explicit focus restoration after the dialog closed, so focus could end up elsewhere.

## Solution

1. **`src/fdeletedlg.pas`**: Use `frmMain` as the dialog owner instead of `Application`, so the LCL restores focus to the main form when the modal closes.
2. **`src/umaincommands.pas`**: Explicitly restore focus to the active file panel after the user confirms deletion.

I built the app and tested it locally, all good.

Fixes #2160


alexx2000:
It is better to pass owner as first argument in `ShowDeleteDialog` and use it (like done in other dialogs):
```
function ShowDeleteDialog(TheOwner: TComponent; const Message: String; FileSource: IFileSource;
  out QueueId: TOperationsManagerQueueIdentifier): Boolean;
begin
  with TfrmDeleteDlg.Create(TheOwner, FileSource) do
```

arsinclair:
Sounds good, I have made the change.

I've also noticed that I committed a build artifact by mistake, not sure why it wasn't ignored. I removed it from the PR and added the ignore statement to the _.gitignore_ file.

Issue #2762, opened 2026-02-20
Synchronize Directories: Incorrect file pairing and missing files with Emoji filenames


axis6404:
### Version

Environment:
debian13 trixie
ii  doublecmd-common  1.1.22-1  all    twin-panel (commander-style) file manager
ii  doublecmd-plugins 1.1.22-1  amd64  twin-panel (commander-style) file manager (plugins)
ii  doublecmd-qt      1.1.22-1  amd64  twin-panel (commander-style) file manager (Qt5)

### Description

7z file:
https://drive.proton.me/urls/CT50A3MS1M#NV6bSEDm5bpo

Description:
The "Synchronize Directories" tool fails to correctly pair files when filenames contain Emojis. The comparison logic misaligns the file list, causing it to compare two entirely different files as if they were a pair.
Steps to Reproduce:
Extract the attached .7z file (contains filenames with Emojis).
Duplicate the folder to create an identical copy in another directory.
Open "Synchronize Directories" and click "Compare" to compare the two folders.
Observe the results and double-click an entry to verify the actual files being compared.

Actual Result:
Missing Files: Several files with Emoji names are completely missing from the result list.
Incorrect Pairing: In the sync list, the tool pairs two different filenames together as if they were the same file.
Verification (Hex Editor): When you double-click such an entry to open the Comparison Viewer, the filenames displayed at the top of the left and right panes are different. This confirms that the tool is comparing unrelated files due to an indexing error.

Expected Result:
All files with Emoji names must be correctly listed.
The tool must only pair and compare files that have the exact same name in their respective directories.
The Comparison Viewer should show identical filenames for both sides.

axis6404:
![Image](https://github.com/user-attachments/assets/0eba5faf-00af-4691-bd2b-09d489ba59a0)

axis6404:
Additional observation:
I haven't tested it extensively, but it seems this issue might only occur with colored Emojis. Monochrome symbols or traditional emoticons might not trigger this behavior.

ags1234:
Too much shit for Doublecmd to handle in a simple synchronize :))

alexx2000:
I cannot reproduce (DC 1.1.32):

<img width="1236" height="720" alt="Image" src="https://github.com/user-attachments/assets/e9c9d71d-7f81-45b9-b2f3-c4d649e2b351" />

axis6404:
[...]

PR #2765, opened 2026-02-22
Fix emoji comparison in TStringListEx #2762

axis6404:
Fix emoji comparison in TStringListEx #2762


alexx2000:
CompareStr/CompareText cannot be used here. They does not support Unicode characters, they compare strings as byte arrays. Maybe UnicodeCompareStr/UnicodeCompareText can be used here instead. Can you test it?

alexx2000:
I applied a more correct fix based on UnicodeCompareStr/UnicodeCompareText.

PR #2782, opened 2026-03-13
Add Cmd+Left/Right commands to open cursor directory in other panel

martinnigel:
Introduce cm_OpenCursorDirInLeftPanel/cm_OpenCursorDirInRightPanel that reuse DoTransferPath after verifying the cursor item is a directory and keep focus in the source panel.

Create runtime actions for those commands, bump hkVersion, and register macOS “Cmd+Left”/“Cmd+Right” hotkeys in the Files Panel context.

This is standard useful shortcut in Total Commander that I am very missing in Double Commander.

### DIFF
--- a/components/viewer/viewercontrol.pas
+++ b/components/viewer/viewercontrol.pas
@@ -4,7 +4,7 @@
     Show file in the text, bin, hex or dec mode
 
     Copyright (C) 2004 Radek Cervinka (radek.cervinka@centrum.cz)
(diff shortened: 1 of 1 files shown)


alexx2000:
DC already has such commands [cm_TransferLeft](https://doublecmd.github.io/doc/en/cmds.html#cm_TransferLeft) (Ctrl+Left) and [cm_TransferRight](https://doublecmd.github.io/doc/en/cmds.html#cm_TransferRight) (Ctrl+Right).

PR #2794, opened 2026-03-21
ADD: Persistent per-panel file mask filter

TheNicker:
Adds a file mask filter that stays active while navigating a panel, following **Total Commander’s** workflow. Left and right panels can use different masks, remember their last-used masks independently, and combine persistent filtering with quick filter.

**Status:** Ready for review.

## Using the filter

Double Commander already provides quick filtering and a mask input dialog. This change adds a persistent “show matching files” workflow through three grouped entries under **Show**:

| Menu entry | Default shortcut | Effect on the active panel |
|---|---|---|
| All files | Ctrl+F10 | Clears the persistent filter |
| Last used | Ctrl+F11 | Reapplies the panel’s last-used mask |
| Custom... | Ctrl+F12 | Opens the existing mask input dialog |

For example, apply `*.exe;*.dll` to one panel and navigate between directories: that panel continues showing matching files. The other panel keeps its own filtering state. Directories remain available for navigation because the persistent mask filters files only.

The **Last used** entry displays the stored mask, shortened when necessary. It is disabled until a mask is available. Shortcuts appear alongside the menu entries.

### Working with quick filter

Persistent filtering and quick filtering have separate state. The persistent mask remains active across navigation; quick filter remains a temporary way to narrow the list.

When both are active, files must pass both filters. For example, a persistent mask of `*.pas` and a quick filter of `test*` narrow the list to matching Pascal files.

Clearing either filter leaves the other active. **All files** therefore removes only the persistent restriction; quick filtering and existing visibility rules can still hide files.

### Seeing the active filter

The panel header appends the persistent mask to the displayed path (similar to **Total Commander**):

```text
C:\Windows\System32\*.exe;*.dll
```

The suffix is display-only. Editing the path edits the real directory path.

The filter label shows whichever filters are active. When both are active, it combines them:

```text
(Filter: *.pas && test*)
```

## Custom commands and bindings

The feature exposes one parameterized command, `cm_PersistentViewFilter`, with `action` and `mask` parameters:

| Parameters | Behavior |
|---|---|
| `action=apply mask=...` | Applies the mask immediately |
| `action=clear` | Clears the persistent filter |
| `action=last` | Reapplies the active panel’s last-used mask |
| `action=dialog` | Opens the mask input dialog |
| `action=dialog mask=...` | Opens the dialog with the supplied mask prefilled |

Omitting `action` defaults to `dialog`. The dialog chooses its initial mask in this order:

1. An explicitly supplied mask.
2. The panel’s current persistent mask.
3. The panel’s last-used persistent mask.

This supports custom bindings with predefined masks. For example, bind **Ctrl+Alt+F10** to `cm_PersistentViewFilter` with:

```text
action=apply
mask=*.png;*.jpg
```

### Panel state and matching

The file view layer owns persistent filter application and clearing, keeping that behavior alongside panel state and navigation.

Persistent and quick filters retain separate text and options. Full list rebuilding and incremental refresh use shared matching logic to apply both consistently. Existing search-template-based mask support is retained.

Persistent masks use explicit mask syntax. Quick filtering keeps its existing filter preparation behavior.

### History and persistence

Left and right panels maintain separate mask histories. When file-mask history saving is enabled, these are stored in `history.xml` under:

```text
PersistentViewFilter
├── Left
└── Right
```

This stores masks for later reuse. The filter’s persistence during navigation and the saved mask history are separate behaviors.

### Commands and display

`cm_PersistentViewFilter` is the public command for custom bindings. Internal menu wrappers dispatch its clear, last-used, and dialog actions.

The header displays the persistent mask, while the filter label displays the combined filtering state. Menu captions, checked states, and shortcuts update to reflect the active panel.

### Scope

The workflow follows Total Commander’s persistent panel filtering, using explicit masks.

It does not add **Show programs**. Defining “program” would require platform-specific decisions about executables and scripts; explicit masks provide a consistent cross-platform approach.

### Changed files

- **Menus and commands:** `src/fmain.lfm`, `src/fmain.pas`, `src/umaincommands.pas`
- **State and matching:** `src/fileviews/ufileview.pas`, `src/fileviews/uorderedfileview.pas`, `src/fileviews/ufileviewworker.pas`
- **Panel display:** `src/fileviews/ufileviewwithpanels.pas`, `src/fileviews/ufileviewheader.pas`
- **History, shortcuts, and strings:** `src/uglobs.pas`, `src/ulng.pas`

### DIFF
--- a/src/fileviews/ufileview.pas
+++ b/src/fileviews/ufileview.pas
@@ -139,6 +139,7 @@ type
     }
     FRequestedActiveFile: String;
     FFilterOptions: TQuickSearchOptions;
+    FQuickFilterOptions: TQuickSearchOptions;
     FWatchPath: String;
     FLastMark: String;
     FLastMarkCaseSensitive: Boolean;
@@ -230,6 +231,7 @@ type
   protected
     FFlatView: Boolean;
     FFileFilter: String;
+    FQuickFilter: String;
     FAllDisplayFiles: TDisplayFiles;    //<en List of all files that can be displayed
     FFiles: TDisplayFiles;              //<en List of displayed files (filtered)
     FSavedSelection: TStringListEx;
@@ -484,6 +486,10 @@ type
     procedure ChangePathToChild(const aFile: TFile); virtual;
 
     procedure ExecuteCommand(CommandName: String; const Params: array of String); virtual;
+    procedure ApplyPersistentViewFilter(const AMask: String); virtual;
+    procedure ClearPersistentViewFilter; virtual;
+    function GetPersistentViewFilterHistory: TStringListEx; virtual;
+    function GetLastPersistentViewFilter: String; virtual;
 
     {en
        Returns @true if at least one file is somehow selected.
@@ -525,6 +531,7 @@ type
 
     procedure SetDragCursor(Shift: TShiftState); virtual; abstract;
     procedure SetFileFilter(NewFilter: String; NewFilterOptions: TQuickSearchOptions);
+    procedure SetQuickFilter(NewFilter: String; NewFilterOptions: TQuickSearchOptions);
     procedure JustForColorPreviewSetActiveState(bActive: Boolean);
 
     property CurrentAddress: String read GetCurrentAddress;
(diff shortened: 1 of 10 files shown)


TheNicker:
@rich2014 , @alexx2000 
Bump.

TheNicker:
Update:  
* Rebased on top of master 2026-09-13.    
* Reduced added net lines of code and optimized code.   
* Checked manually known edge cases and backward compatibility.  
* Added TODOs for fixing encountered possibly pre-existing sub-optimal implementations.

PR #2796, opened 2026-03-21
UPD: Move DirectoryHotList to localconfig.xml

TheNicker:
#  Move DirectoryHotList to localconfig.xml and introduce machine-specific local configuration

  # Description

  This change separates DirectoryHotList from doublecmd.xml and stores it in a new localconfig.xml file.

  The motivation is that filesystem structure differs between systems. A shared doublecmd.xml works well for portable/common settings, but DirectoryHotList often contains machine-specific paths that should not
  be synchronized across different environments.

  This PR also opens a more general-purpose local configuration layer. localconfig.xml is not introduced as a hotlist-only file, it is intended to be the place for settings that are inherently system-specific
  and should remain outside the shared main configuration.

  # What Changed

  - Added a new machine-specific configuration file: localconfig.xml
  - Moved DirectoryHotList serialization/deserialization to localconfig.xml
  - Kept the XML structure under:
```
    <doublecmd>
      <DirectoryHotList>
        ...
      </DirectoryHotList>
    </doublecmd>
```
  - Added migration logic for legacy configurations
  - Bumped config version to trigger one-time migration behavior

  # Migration Behavior

  - Existing configurations still load correctly
  - If localconfig.xml does not yet contain DirectoryHotList, Double Commander falls back to the legacy node in doublecmd.xml
  - For configs older than version 16, the legacy DirectoryHotList node is removed from the in-memory main config after it is migrated, so it disappears from doublecmd.xml on the next save

  # Why This Is Useful

  - Shared configuration becomes cleaner and more portable between systems
  - Machine-specific path data is isolated
  - Establishes a reusable mechanism for future local-only settings

  # Notes

  - doublecmd.xml remains the shared configuration file
  - localconfig.xml is the new machine-specific companion configuration file

### DIFF
--- a/src/uglobs.pas
+++ b/src/uglobs.pas
@@ -231,7 +231,8 @@ const
   // 13 -  Replace Configuration/UseConfigInProgramDir by doublecmd.inf
   // 14 -  Move some colors to colors.json
   // 15 -  Move custom columns colors to colors.json
-  ConfigVersion = 15;
+  // 16 -  Move DirectoryHotList to localconfig.xml
+  ConfigVersion = 16;
 
   COLORS_JSON = 'colors.json';
 
@@ -996,6 +997,33 @@ begin
   end;
 end;
 
+function LoadLocalConfig(var {%H-}ErrorMessage: String): Boolean;
+var
+  Root: TXmlNode;
+  LocalConfig: TXmlConfig;
+  LocalConfigFileName: String;
+begin
+  LocalConfigFileName := gpCfgDir + gcfLocalConfig;
+
+  if mbFileExists(LocalConfigFileName) then
+  begin
+    LocalConfig:= TXmlConfig.Create(LocalConfigFileName);
+    try
+      LocalConfig.Load;
+      Root:= LocalConfig.RootNode;
+      if Assigned(LocalConfig.FindNode(Root, cSectionOfHotDir)) then
+      begin
+        gDirectoryHotlist.LoadFromXml(LocalConfig, Root);
+        Exit(True);
+      end;
+    finally
+      LocalConfig.Free;
+    end;
+  end;
+
+  Result:= True;
+end;
+
 procedure SaveHistoryConfig;
 var
   Root: TXmlNode;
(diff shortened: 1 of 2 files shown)


TheNicker:
@alexx2000 , what lingers this PR, Is the functionality is unclear, non-trivial , or requires further functional design discussion ? 
And on that note, who else is eligible for reviewing it ?

src/uglobs.pas
```diff
@@ -1049,6 +1085,25 @@ procedure SaveHistoryConfig;
...
+var
+  Root: TXmlNode;
+  LocalConfig: TXmlConfig;
+  LocalConfigFileName: String;
+begin
+  LocalConfigFileName := gpCfgDir + gcfLocalConfig;
+  LocalConfig:= TXmlConfig.Create(LocalConfigFileName);
+  try
+    if mbFileExists(LocalConfigFileName) then
+      LocalConfig.Load;
```
alexx2000: What is this code for? We overwrite everything below.

src/uglobs.pas
```diff
@@ -996,6 +997,41 @@ function LoadHistoryConfig(var {%H-}ErrorMessage: String): Boolean;
...
+      begin
+        gDirectoryHotlist.LoadFromXml(LocalConfig, Root);
+        Exit(True);
+      end;
+    finally
+      LocalConfig.Free;
+    end;
+  end;
+
+  // Backward compatibility: read the legacy node from doublecmd.xml
```
alexx2000: I think this part of code is better to move to `LoadXmlConfig` as it was before, but like this:

```
if LoadedConfigVersion < 16 then
begin
  gDirectoryHotlist.LoadFromXML(gConfig, Root);
  gConfig.DeleteNode(Root, cSectionOfHotDir);
end;
```

antodc:
> # Move DirectoryHotList to localconfig.xml and introduce machine-specific local configuration
> 
>     * doublecmd.xml remains the shared configuration file
> 
>     * localconfig.xml is the new machine-specific companion configuration file

I think this is a nice improvement for the usability of DC across different systems. If I may add on this, there are also other parts of the Options where directories appear (Configuration, Log, Plugins, maybe others). I think it would be nice if in the future also those parts could be "moved" to the localconfig file.

src/uglobs.pas
TheNicker: Done. I moved the legacy DirectoryHotList migration into LoadXmlConfig, next to the other ConfigVersion-based migrations, and removed the fallback/migration logic from LoadLocalConfig. LoadLocalConfig now only reads DirectoryHotList from localconfig.xml if that section exists.

src/uglobs.pas
TheNicker: Done. Assuming SaveLocalConfig is intended to save all localconfig.xml sections, future local-only sections should be written there explicitly, so loading the previous file first is unnecessary.

TheNicker:
> I think this is a nice improvement for the usability of DC across different systems. If I may add on this, there are also other parts of the Options where directories appear (Configuration, Log, Plugins, maybe others). I think it would be nice if in the future also those parts could be "moved" to the localconfig file.

 @antodc, yes, I agree there are likely more machine-specific settings. That is the broader idea behind this PR.

  Since this approach is already accepted, I suggest using this PR as a stability check for the localconfig.xml mechanism and
  migration procedure. Could you please open an issue listing the other machine-specific settings that should be considered
  for migration? Then, once this proves stable, we can review the candidates and move the remaining local machine-specific
  configuration in a single configuration version bump.

PR #2837, opened 2026-04-19
Put e.g. 5 single-line JSON files of 24MB in a single folder to cause…

koenvandesande:
… viewer to hang for seconds

### DIFF
--- a/src/fviewer.pas
+++ b/src/fviewer.pas
@@ -2960,7 +2960,7 @@ begin
   begin
     PushPop(FElevate);
     try
-      if FileHasLongLines(sFileName, 32768) then
+      if FileHasLongLines(sFileName, gMaxTextWidth) then
         Result := False;
     finally
       PushPop(FElevate);
--- a/src/fviewer.pas
+++ b/src/fviewer.pas
@@ -2872,6 +2872,36 @@ begin
   end;
 end;
 
+// Returns True if the file has lines longer than MaxLen bytes.
+// Only reads the first MaxLen+1 bytes to check.
+function FileHasLongLines(const sFileName: String; MaxLen: SizeInt): Boolean;
+var
+  CheckLen, J: SizeInt;
+  CheckBuf: AnsiString;
+begin
+  Result := False;
+  try
+    with TFileStreamUAC.Create(sFileName, fmOpenRead or fmShareDenyNone) do
+    try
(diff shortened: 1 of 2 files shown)


koenvandesande:
[sbc_case_builder.json](https://github.com/user-attachments/files/26876365/sbc_case_builder.json)

Put 5 or 6 copies of this file into a single folder. Open the built-in viewer. The hang is more than a second per file. This fixes the slowdown in the thumbnail thread and in the main viewer by falling back to non-syntax highlighting for very long lines.

Skif-off:
What about `gMaxTextWidth` aka [MaxTextWidth](https://doublecmd.github.io/doc/en/configxml.html#MaxTextWidth)?

koenvandesande:
> What about `gMaxTextWidth` aka [MaxTextWidth](https://doublecmd.github.io/doc/en/configxml.html#MaxTextWidth)?

Good point, I replaced the constants with this variable instead.

Skif-off:
Maybe the best way is another parameter as an addition to [MaxCodeSize](https://doublecmd.github.io/doc/en/configxml.html#MaxCodeSize)?
I.e., any plain text and the text that will be processed by the parser for syntax highlighting are very different entities by necessary resources (CPU, RAM). I'm not sure if I've managed to express my thought correctly :))

koenvandesande:
With a maximum line length of MaxTextWidth, we can determine if the file has (too long) lines from reading just MaxTextWidth bytes. So that setting is perfect for that.
For the preview, it now reads 8x MaxTextWidth, which is typically 8KB. That is superfast. Before, it would read the entire 24MB file.

The MaxCodeSize is in Megabytes, that would be a much too large value to use.

Skif-off:
For example, user can intentionally decrease the `MaxTextWidth` value to use this as an alternative to the "View > Wrap text" option. Then this condition
```pascal
if FileHasLongLines(sFileName, gMaxTextWidth) then
```
will give a slightly unexpected result.

I don't know, just a thought out loud :)

koenvandesande:
With that thought, I think it is best to go to a conservative constant value of 8192: if a file doesn't have a first line break within that, then we should not use SynEdit syntax highlighting.

For the thumbnails, the first 8192 should be enough to render the preview of the first lines (and otherwise you see a bit less in the preview).

I don't think this behavior should be configurable, because most users won't understand what it is about.

PR #2875, opened 2026-05-22
ADD: SFTP/FTP - preserve file permissions on transfer (when supported by the server)

emilioheredia-source:
When transferring files via SFTP on Linux/Unix, file permissions (mode bits) are not preserved:
- **Upload**: the remote file gets the server's default umask permissions instead of the local file's permissions
- **Download**: the local file loses its permissions — for example, an executable downloaded from the server loses the execute bit and cannot be run without manually restoring permissions with `chmod`

Most SFTP clients (FileZilla, Cyberduck, rsync, scp) preserve permissions by default. DC already handles permissions correctly for local copy/move and for tar/zip archive extraction — SFTP was the only gap.

### Fix

In `TSftpSend` (`plugins/wfx/ftp/src/sftp/sftpsend.pas`):

- **`StoreFile` (upload)**: after the transfer completes, calls `libssh2_sftp_setstat` with the local file's `st_mode` (permissions) and `st_uid`/`st_gid` (ownership). Ownership is best-effort — SFTP servers silently ignore it for non-root users, which is standard behaviour (same as `rsync`).
- **`RetrieveFile` (download)**: after the transfer completes, calls `libssh2_sftp_stat` to read the remote file attributes, then `FpChmod` to apply the mode bits to the local file. `FpChown` is also called for ownership, again best-effort.

### Platform impact

All new code is inside `{$IFDEF UNIX}` blocks. On Windows the file compiles identically to before — no behaviour change, no new symbols referenced. The `BaseUnix` unit (FPC standard library) is added to the `uses` clause under the same guard.

### Change

One file modified: `plugins/wfx/ftp/src/sftp/sftpsend.pas`, 38 lines added.

### DIFF
--- a/plugins/wfx/ftp/src/ftpadv.pas
+++ b/plugins/wfx/ftp/src/ftpadv.pas
@@ -127,6 +127,7 @@ type
     function FileProperties(const FileName: String): Boolean; virtual;
     function CopyFile(const OldName, NewName: String): Boolean; virtual;
     function ChangeMode(const FileName, Mode: String): Boolean; virtual;
+    function GetRemoteMode(const FileName: String; out Mode: TFileAttrs): Boolean; virtual;
     function List(Directory: String; NameList: Boolean): Boolean; override;
     function StoreFile(const FileName: string; Restore: Boolean): Boolean; override;
     function ExecuteCommand(const Command: String; const Directory: String = ''): Boolean; virtual;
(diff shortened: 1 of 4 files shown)


Skif-off:
What about FTP?

emilioheredia-source:
Done — I applied the same idea to FTP transfers as well (included in this PR).

Fair warning though: FTP is an older protocol and permission support varies heavily by server implementation. The FTP side uses SITE CHMOD on upload and MLST unix.mode on download, both of which are optional extensions. It should degrade gracefully on servers that don't support them — the transfer completes normally, just without the permission fix applied.

I only tested it against pyftpdlib (Python package), which is handy for a quick local FTP server. Haven't tested against vsftpd, proftpd, or pure-ftpd yet, though the code follows their documented behaviour for these commands.

Cheers.

PR #2892, opened 2026-05-26
FIX: sync compare — treat equivalent symlink targets as equal

emilioheredia-source:
## Problem

When synchronizing with symlink handling set to skip/follow-as-link, Double Commander may recreate links with a different textual target representation (for example, a shorter relative path on one side and a longer rebased relative path on the other).

Even when both links resolve to the same real target, sync compare marked them as different because it used mtime + size only. On Linux, symlink size is the link text length, so equivalent links with different textual forms appear unequal.

## Example

- Left link target: ../java.base/ADDITIONAL_LICENSE_INFO (size 36)
- Right link target: ../../../../Programs/pycharm-professional/jbr/legal/java.base/ADDITIONAL_LICENSE_INFO (size 85)

Both resolve to the same file, but were flagged different.

## Fix

In TFileSyncRec.UpdateState (fsyncdirsdlg.pas), add symlink-aware comparison:

- If both sides are links and link text matches, treat as equal.
- Otherwise resolve both link targets to absolute paths and compare resolved targets.
- Keep the existing mtime logic unchanged for regular files.

This preserves original sync intent (semantic equality) and avoids false positives for equivalent links.

## Testing

- Sync two trees containing equivalent symlinks with different textual target forms.
- Verify they are no longer listed as different after sync.
- Verify regular file compare behavior remains unchanged.

### DIFF
--- a/src/fsyncdirsdlg.pas
+++ b/src/fsyncdirsdlg.pas
@@ -543,6 +543,30 @@ end;
 procedure TFileSyncRec.UpdateState(ignoreDate: Boolean);
 var
   FileTimeDiff: Integer;
+
+  function AreEquivalentLinks: Boolean;
+  var
+    LeftTarget, RightTarget: String;
+    LeftResolved, RightResolved: String;
+  begin
+    Result := False;
+
+    if not (FFileL.IsLink and FFileR.IsLink) then Exit;
+
+    LeftTarget := FFileL.LinkProperty.LinkTo;
+    RightTarget := FFileR.LinkProperty.LinkTo;
+
+    // Fast path: identical link text means semantically identical link.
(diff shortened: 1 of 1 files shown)


emilioheredia-source:
Context: this UI-only PR complements the earlier sync correctness/performance fixes already opened:\n\n- #2887 (symlink follow/skip choice persistence)\n- #2888 (symlink timestamp preservation on Linux)\n- #2889 (sync progress pump throttling / O(n^2) slowdown)\n- #2890 (asymmetric mode notEq handling)\n- #2892 (equivalent symlink target comparison)\n\nThis PR (#2893) only aligns sync progress bar rendering with the regular file-op dialog (smooth mode), with no behavior changes in sync logic.

Issue #2891, opened 2026-05-26
Command-line local drive path opens network root when active pane is on UNC share


TheNicker:
### Version

Double Commander
Version: 1.3.0 alpha
Revision: 12980
Commit: 246f41067
Build: 2026/05/26
Lazarus: 4.6.0.0
Free Pascal: 3.2.2
Platform: x86_64-Win64-win32/win64
System: Windows 11 25H2 x86_64

### Description

When Double Commander is already running and the active pane is currently opened on a UNC network share, launching Double Commander again from the command line with a local drive path does not open the requested local path.

Instead, the existing Double Commander window shows the network root with all discovered network names.

Steps to reproduce:

1. Open Double Commander and leave the window running.
2. In the active pane, navigate to a UNC network share, for example:

   ```text
   \\server-1\c$
   ```

3. From a command prompt or PowerShell, run:

   ```text
   doublecmd c:/
   ```

Expected result:

The active pane in the existing Double Commander window should navigate to:

```text
C:\
```

Actual result:

The active pane shows the network root with all discovered network names instead of opening `C:\`.

For example, if the active pane is at:

```text
\\server-1\c$
```

then running:

```text
doublecmd c:/
```

opens the discovered network locations view instead of treating `c:/` as the absolute local drive path.

PR #2902, opened 2026-05-31
FIX: Navigate to correct file source when opening path from command line

TheNicker:
## Summary

Solves issue #2891.

- Routes command-line paths through file-source selection when reusing an existing panel, so local paths are not forced through the currently active network file source.
- Captures the active `FileView` before calling `ChooseFileSource`, so selecting a file after navigation targets the same view that was navigated.
- Narrows Windows network root detection so local drive paths like `C:\` are not treated as network browsing paths.

## Bug Fix

- Fixes opening a local drive path from a second instance when the target pane is currently on a UNC/network location.
- Fixes command-line file paths by switching to the correct directory/file source before selecting the file.
- Avoids collapsing UNC command-line paths through deepest-existing-local-path fallback.

## Solved Cases

- Existing instance active on a UNC/network pane, then running `doublecmd C:\` opens the local `C:\` path instead of the network root.
- Existing instance active on a UNC/network pane, then running `doublecmd C:\path\file.ext` opens the local directory and selects `file.ext`.
- Existing instance receiving a UNC path like `doublecmd \\server\share` keeps the network path route instead of treating it as a local-drive fallback case.

### DIFF
--- a/src/filesources/winnet/uwinnetfilesource.pas
+++ b/src/filesources/winnet/uwinnetfilesource.pas
@@ -176,7 +176,9 @@ end;
 
 function TWinNetFileSource.IsNetworkPath(const Path: String): Boolean;
 begin
-  Result:= (NumCountChars(PathDelim, ExcludeTrailingPathDelimiter(Path)) < 3);
+  Result:= (Path = PathDelim) or
+           ((Pos(PathDelim + PathDelim, Path) = 1) and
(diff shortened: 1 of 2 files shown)


TheNicker:
Can we move forward with it , see #2891. this PR solves the use cases described in the issue cases

alexx2000:
I applied a modified version. It should work with other FileSources too.

TheNicker:
> I applied a modified version. It should work with other FileSources too.

Thanks

PR #2917, opened 2026-06-06
FIX: SFTP - preserve symlinks when copying local folder to SFTP

emilioheredia-source:
When copying a local directory tree to an SFTP destination, symlinks were silently converted to regular files (if the target existed) or caused copy errors (if the target was missing or broken).

**Root cause — three layered bugs:**

1. `uwfxplugincopyinoperation.pas`: the WFX copy-in tree builder used `fsooslFollow`, which resolved symlinks to their real targets before `StoreFile` was ever called. The plugin never saw the symlink itself.

2. `uwfxpluginutil.pas / ProcessLink`: had no `else` branch for the no-subnode case (which is the normal case when not following links), leaving `Result` uninitialised — an existing latent bug.

3. `sftpsend.pas / StoreFile`: opened the local path unconditionally with `TFileStreamEx`, which follows symlinks on the OS level — so even a symlink to a real file would be uploaded as a plain copy.

**Fix:**

- `sftpsend.pas`: detect symlinks via `fpLStat` before opening the file. If the source is a symlink, read its target with `fpReadLink` and call `libssh2_sftp_symlink` to recreate it on the remote (rsync-like behaviour). Only affects Unix builds (`{ UNIX}`).
- `uwfxplugincopyinoperation.pas`: switch `SymLinkOption` to `fsooslDontFollow` so the symlink path itself reaches `StoreFile` intact.
- `uwfxpluginutil.pas / ProcessLink`: add `else` branch for the no-subnode case to route the symlink node to `ProcessFile`, fixing the uninitialised `Result`.

**Behaviour after fix:** symlinks in a copied folder are recreated as symlinks on the SFTP server (relative or absolute, exactly as on the source), matching rsync's default.

### DIFF
--- a/plugins/wfx/ftp/src/sftp/sftpsend.pas
+++ b/plugins/wfx/ftp/src/sftp/sftpsend.pas
@@ -228,11 +228,38 @@ var
   TotalBytesToWrite: Int64 = 0;
   TargetHandle: PLIBSSH2_SFTP_HANDLE = nil;
   Flags: cint = LIBSSH2_FXF_CREAT or LIBSSH2_FXF_WRITE;
+{$IFDEF UNIX}
+  LocalStat: BaseUnix.TStat;
+  LinkTarget: String;
+{$ENDIF}
 begin
   if FCopySCP then begin
     Result:= inherited StoreFile(FileName, Restore);
     Exit;
   end;
+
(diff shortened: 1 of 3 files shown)


emilioheredia-source:
Closing to review before submitting.

PR #2982, opened 2026-07-13
Internal viewer: right-align text for RTL (Hebrew) files

meirtsvi:
## Summary

When the internal viewer (F3) opens a text file containing Hebrew characters, the document is now displayed right-to-left: lines are right-aligned in text/wrap/book modes, matching how RTL documents are normally read.

- Detection counts characters in the first 4kB of the *decoded* text (same approach as `DetectEncoding`) and right-aligns only when characters from the Hebrew Unicode block (U+0590–U+05FF) outnumber the rest. It runs on file load and re-runs whenever the encoding is changed, so switching encodings updates the alignment correctly.
- The selection highlight and the caret are anchored to the same right edge as the rendered line, so selecting text in right-aligned lines highlights the correct region.
- Click-to-position mapping (`XYPos2Adr`) is mirrored for RTL lines: the distance from the right edge corresponds to the accumulated logical text width, so mouse selection lands on the intended characters (e.g. double-clicking the visually rightmost word selects the logically first word).
- Hex/bin/dec modes and LTR files are unaffected.

## Testing

Built and verified on macOS (Cocoa, aarch64, Lazarus main):
- Hebrew UTF-8 file opens right-aligned with correct RTL rendering; English file remains left-aligned.
- Switching the encoding to ANSI (mojibake, no Hebrew after decode) correctly reverts to left alignment; switching back restores RTL.
- Double-clicking the visually rightmost Hebrew word selects the logically first word, with the highlight drawn in place.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/viewer/viewercontrol.pas
+++ b/components/viewer/viewercontrol.pas
@@ -225,6 +225,7 @@ type
     FMaxTextWidth:       Integer; // maximum of chars on one line unwrapped text (max 16384)
     FExtraLineSpacing:   Integer;
     FLeftMargin:         Integer;
+    FRightToLeft:        Boolean; // text contains RTL characters (e.g. Hebrew) - lines are right-aligned
     FOnGuessEncoding:    TGuessEncodingEvent;
     FOnFileOpen:         TFileOpenEvent;
     FCaretVisible:       Boolean;
@@ -254,6 +255,7 @@ type
     procedure SetTabSpaces(const AValue: Integer);
     procedure SetShowCaret(AValue: Boolean);
     procedure SetCaretPos(AValue: PtrInt);
+    function DetectRightToLeft: Boolean;
 
     {en
        Returns how many lines (given current FTextHeight) will fit into the window.
@@ -1871,11 +1873,12 @@ begin
   FBOMLength := 0;
   FBlockBeg  := 0;
   FBlockEnd  := 0;
+  FRightToLeft := False;
 end;
 
 procedure TViewerControl.WriteText;
 var
-  yIndex, xIndex, w, i: Integer;
+  yIndex, xIndex, w, i, X: Integer;
   LineStart, iPos: PtrInt;
   CharLenInBytes: Integer;
   DataLength: PtrInt;
@@ -1888,6 +1891,14 @@ var
       FCaretPoint.X:= X;
       FCaretPoint.Y:= Y;
     end;
+  end;
+
+  function EmptyLineX: Integer;
+  begin
+    if FRightToLeft then
+      Result := ClientWidth - FLeftMargin
+    else
+      Result := 0;
   end;
 
 begin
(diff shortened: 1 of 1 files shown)


Skif-off:
Can't enable RTL support in SynEdit?

meirtsvi:
> Can't enable RTL support in SynEdit?

It does RTL but not right-aligning the text, which makes it irrelevant to view Hebrew text properly.

meirtsvi:
Anyone?

Skif-off:
"RTL" is usually Arabic, Hebrew and Syriac or Arabic and Hebrew: is this a problem? (I use a simple alphabetical sort if that's a problem.)
AFAIR it's one group of languages (Semitic languages), so just need to expand the range of Unicode characters?

If I understand correctly, the symbols are checked in any case. But, for example, if `DetectEncoding` returned ISO 8859-8 (OEM 862, OEM 864, ISO 8859-6, Win-1256, MacArabic and so on), then this is not necessary?
Or is the additional time for verification small and in theory not critical?

meirtsvi:
My fix aligns the text to the right. The text shows from right to left, but it's not aligned to the right corner. With my fix, it does:
align_right

meirtsvi:
@Skif-off can this be approved?

Skif-off:
> can this be approved?

I don't know, I just saw something I didn't understand and asked:

For some reason unknown to me, you have limited yourself to only a part of the common languages that use right alignment.

Your method with `DetectRightToLeft` does not seem appropriate. For example, I copied a few lines from Wikipedia (I am not familiar with Semitic languages), this is an English text that contains several words in Hebrew: [example.txt](https://github.com/user-attachments/files/30223070/example.txt)
This text should not be right-aligned, but...

P.S. The inevitable additional conclusion: this feature should be optional (maybe with internal command).

meirjfrog:
@alexx2000 can you please review?

alexx2000:
I agree with @Skif-off:

> Your method with DetectRightToLeft does not seem appropriate.

Maybe `DetectRightToLeft` should count number of characters (Hebrew/non-Hebrew) and return `True` when Hebrew characters count is greater.

meirtsvi:
> I agree with @Skif-off:
> 
> > Your method with DetectRightToLeft does not seem appropriate.
> 
> Maybe `DetectRightToLeft` should count number of characters (Hebrew/non-Hebrew) and return `True` when Hebrew characters count is greater.

@alexx2000 Done

meirtsvi:
@alexx2000 ping :)

Skif-off:
Does analysis significantly affect file opening speed?
Perhaps it would be better to make this feature optional, just in case?

meirtsvi:
The code scans the first 4k characters, so it's pretty fast. The response time is immediate, tested on smal/medium/large.

Skif-off:
There's also encoding detection and a code viewing mode. I also thought about the possibility of adding other RTL languages. But milliseconds and CPU cycles have habit to accumulate...
Perhaps I am being overly cautious :))

meirtsvi:
The action of viewing a file is manual, and the time it takes for the viewer to show up, even for huge files, is instantaneous.
https://github.com/user-attachments/assets/aef1b12c-cab8-449a-8b2f-57e75d771bef

meirtsvi:
@alexx2000 ping :)

meirjfrog:
@alexx2000 ping, I've added a movie above to prove that no real performance issue exist even with very large files

PR #3017, opened 2026-08-03
FIX: keep saved path of inactive tabs on startup

### LINKED ISSUE #2502: Inactive tabs are touched on launch, losing path info for inaccessible targets
### Version

Double Commander
Version: 1.1.28 gamma
Revision: 787
Commit: b204fbb61
Build: 2025/08/10
Lazarus: 4.3.0.0
Free Pascal: 3.2.2
Platform: x86_64-Win64-win32/win64
System: Windows 10 Home 22H2 x86_64

### Description

Inactive tabs are touched on launch, losing path info for inaccessible targets.
This is disruptive for workflows using remote/removable destinations.

Don't know if DC tries to pre-load tab contents or just eagerly checks the paths, but keeping the path for each tab until focused is certainly doable and would be greatly appreciated.

ZoomRmc:
`TFileView.LoadConfiguration` corrected the path of every restored tab (`GetDeepestExistingPath` with `mbGetCurrentDir` fallback), so a tab whose target was temporarily unavailable (unplugged drive, offline network share) had its saved path silently rewritten at launch and persisted on exit.

This runs the correction only when the tab is actually loaded (i.e. when it does not have the `fvfDelayLoadingFiles` flag). 

Cache path availability for a '?' tab-title marker that is refreshed on activation, path change, list load and drive-added events.
On activation of a tab with unavailable path an empty list is displayed and `..` will follow through to the first available parent. This prevents silent changing of tab's path.

The marker is controlled via the new default `tb_show_unavailable_marker` option.

Closes #2502

### DIFF
--- a/src/fileviews/ufileview.pas
+++ b/src/fileviews/ufileview.pas
@@ -112,6 +112,7 @@ type
     FFilePropertiesNeeded: TFilePropertiesTypes;
     FFileViewWorkers: TFileViewWorkers;
     FFlags: TFileViewFlags;
+    FPathAvailable: Boolean;     //<en Cached availability of the current path (used by the tab marker)
     FHashedFiles: TBucketList;  //<en Contains pointers to file source files for quick checking if a file object is still valid
     FHashedNames: TStringHashListUtf8;
     FPendingFilesChanges: TFPList;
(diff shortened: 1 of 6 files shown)


ZoomRmc:
Thanks for the merge! This issue's been bugging me for years.

PR #3022, opened 2026-08-04
ADD: Monochrome tray icon support

corpix:
Hey, I have added monochrome icon support.
It could be enabled from options window:
image

Here is how it looks like in awesome-wm:
image

What do you think?


ags1234:
Much of the beauty of this world is in the colors and this is very well supported by the fact that the human eye adapted to distinguish approximately 10 million colors.
Could never figure out myself why would people use monochrome themes combined with monochrome wire-frame icons.

This is a picture from my "modern design" KDE Plasma desktop:


Bland af.

Oh, where are you KDE 3.5? Oh, where are you BeOS?  Ha ha

PR #3031, opened 2026-08-12
fix: restore main window focus after closing delete confirmation

arsinclair:
My earlier focus fix is still present - 3f9e3d3, but it restores focus synchronously immediately after ShowModal returns. It seems that with newer Qt6/LCL versions, native modal window closing and activation events may finish processing afterward, causing the main window to lose focus again.

The solution will queue the focus restoration so it runs after the widgetset has completed closing the dialog. Bring the main window to the front first, then restore focus to the active file panel.

### DIFF
--- a/src/umaincommands.pas
+++ b/src/umaincommands.pas
@@ -58,6 +58,7 @@ type
                                           State: TFileSourceOperationState);
    procedure OnCalcChecksumStateChanged(Operation: TFileSourceOperation;
                                         State: TFileSourceOperationState);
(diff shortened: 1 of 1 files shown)


arsinclair:
Just wondering if this PR slipped off the radar. @alexx2000 @rich2014 could you take a look when you get a minute, please. Much appreciated.

rich2014:
@arsinclair 

hi, i'm not familiar with Qt and can't assess the impact of this change on other platforms, so i can't offer any opinions.

Issue #1561, opened 2024-04-08
Random freeze/beachball when browsing SMB volumes


nitrostudios:
### Version

Double Commander
Version: 1.2.0 alpha
Revision: 11277
Commit: 17c4b7f46
Build date: 2024/03/31
Lazarus: 3.99.0.0
FPC: 3.2.2
Platform: aarch64-Darwin-cocoa
OS version: Mac OS X 12.6.6

### Description

I have a problem with random DC freeze/beachball when browsing the SMB volumes, sometimes more, sometimes less often. It's especially bad when I have opened tab with remote location and there is a copy operation in progress from some other client, then the whole interface freezes until the copy operation is completed. Does not happen in Finder or Forklift 4, but the same thing happens also in Nimble Commander, for example. Anyone else having the same problem?

nitrostudios:
I've done some more testing and the it seems that only the application GUI freezes, because if there is an operation in progress (copy, move), it's still going on in the background, I can see it in the network monitor or disk activity. Can something like that depend on the API - Cocoa VS SwiftUI ? The Finder and Forklift 4 are using the SwiftUI.

alexx2000:
I don't think so. Simply DC works with network folder same way as with local folder. Any network freeze will freeze DC.

alex-s-team:
happens to me as well. can confirm that finder and forklift 4 don't have that problem. can also confirm that a network freeze triggers it - i.e. a directory with a (very) large number of files gets refreshed. maybe the multithreaded refresh / reload of a directory pane is handled differently?

kaitrek:
Related to network freezes, is it possible to run each Pane in its own process so that when right pane is frozen accessing a remote location, the left pane can still browse?

nitrostudios:
Just a small update: I've been in touch with Apple several times and developers are aware of the issue and are currently looking for more information. The problem seems to be with macOS SMB implementation. If there is anyone that is experiencing the same issue, please report it to the Apple support. I've been told that the more people will report the higher chance is for a fix to happen!

chachizhu:
@alexx2000 
Hello
I found that making the following changes on macOS can reduce the likelihood of beachball occurrences:
	1.	In the ```Colors``` configuration, when matching file types, the default subfolder search depth is set to ```all```. This needs to be changed to ```current```.
	2.	In the ```Auto Refresh``` configuration, disable ```Auto Refresh``` for SMB mappings.

alexx2000:
Hi,

[...]

Issue #1266, opened 2023-09-16
macos: DC hangs too often on multiple files ops on external hdd


zxcat:
### Version

Checked:
* latest snapshot: doublecmd-1.2.0-10923.cocoa.x86_64,
* latest beta: doublecmd-1.1.2-71.cocoa.x86_64.
* Earlier has this issue too (some 1.1.0 alpha).

OS: macos 10.15.7.

### Description

I cannot get detailed error log, because DC just hangs and can only be force killed. Furthermore, now it is in some zombie state, so I can't even restart or completely kill it : )

Complete steps to reproduce are unknown. DC hangs when I run several heavy file ops on external HDD and OS do some heavy ops too. E.g. OS cleans up Trash can with 200k files and I try to move several 15Gb+ files from one external HDD to another or traverse dir containing 4k dirs with total 150+k files inside. Or sometimes I just access one file when heavy op happens on the same hdd…

Seems DC somehow locks itself with mutexes and then hangs. Other apps can access HDDs without issue.

Here are apple reports obtained after force kill:

<details><summary>1.2.0 (10923) one</summary>

```
Date/Time:        2023-09-16 16:29:18 +0300
End time:         2023-09-16 16:31:48 +0300
OS Version:       Mac OS X 10.15.7 (Build 19H2026)
Architecture:     x86_64h
Report Version:   29

Data Source:      Stackshots
Shared Cache:     0xc123000 D14D9CAD-E258-3364-A090-5121E756DDB0

Command:          doublecmd
Path:             /Users/USER/Downloads/*/Double Commander.app/Contents/MacOS/doublecmd
Identifier:       com.company.doublecmd
Version:          1.2.0 (10923)
Parent:           launchd [1]
PID:              57682

Event:            hang
Duration:         150.05s
Duration Sampled: 1.09s (process was unresponsive for 149 seconds before sampling)
Steps:            11 (100ms sampling interval)

Hardware model:   iMac20,1
Active cpus:      16
Boot args:        chunklist-security-epoch=0 -chunklist-no-rev2-dev

Time Awake Since Boot: 390000s
Time Since Wake:  210000s


--------------------------------------------------
Timeline format: stacks are sorted chronologically
Use -i and -heavy to re-report with count sorting
--------------------------------------------------


Heaviest stack for the main thread of the target process:
  11  ??? (doublecmd + 11901) [0x1067cce7d]
  11  ??? (doublecmd + 697944) [0x106874658]
[...]

PR #3052, opened 2026-08-23
FIX: avoid macOS FSEvents UI deadlock during stream teardown

EricXh:
Addresses #1561

Related to #1266 and commit 8494d60e8a97c0b81efef3171ce102799060a06a.

## Problem

On macOS, changing directories on an SMB share can freeze the entire Double Commander UI for 5-15 seconds even when raw SMB directory enumeration is fast.

Process samples show the FSEvents worker blocked in:

`FSEventStreamInvalidate -> __CFFileDescriptorDeallocate -> dispatch_sync -> kevent_id`

At the same time, the main thread is waiting for the watcher mutex while adding or removing a watched path.

## Root cause

`TDarwinFSWatcher.start` calls `updateStream` while holding `_lockObject`. `updateStream` may synchronously flush, stop, invalidate, and release the previous FSEvent stream. On SMB, Apple's teardown can block for many seconds, so UI-side `addPath` and `removePath` calls wait for the same mutex and the GUI beachballs.

## Fix

- Copy `_watchPaths` to an immutable `NSArray` snapshot while holding `_lockObject`.
- Release `_lockObject` before rebuilding or closing the FSEvent stream.
- Reacquire the mutex afterward and compare the current watched paths with `_streamPaths`.
- Repeat the rebuild when watched paths changed during the slow teardown.

This keeps the potentially slow Apple call on the watcher thread without holding the mutex needed by the UI. Path changes that arrive during teardown are retried rather than lost.

## Relation to #1266

The fix for #1266 filters subdirectory events unless Flat View is active, reducing unnecessary refresh traffic. This change is complementary: it removes the remaining UI lock coupling when an FSEvent stream itself is synchronously torn down.

## Tests

Environment: macOS 26.6.2 on Apple Silicon, Cocoa, FPC 3.2.2, Lazarus 4.99, TrueNAS SMB share.

- 2,640 automated SMB directory transitions across both panes on the patch-only build, with no freeze or crash.
- 74 UI responsiveness probes completed in 386-514 ms.
- Two samples caught `FSEventStreamInvalidate` blocking the watcher for 17.739 and 8.627 seconds while the main thread stayed responsive in the AppKit event loop.
- 100 concurrent SMB create/remove cycles completed during navigation.
- Auto-refresh detected a 20-file create/remove burst without manual refresh.
- Kqueue count remained exactly 2; no material FSEvents leak was detected.
- Clean shutdown after stress completed in 1.078 seconds without a crash report.
- Rebased onto current master `1c4c3ef0fcdba54d4aef7d8b6832b01e891a2a21`; ARM64 build and `git diff --check` pass.
- Focused current-master rerun: 120 additional SMB transitions, auto-refresh 0 -> 10 -> 0 files, and clean exit in 1.019 seconds.

## Current-master test note

A completely clean current-master build on macOS 26.6.2 also hits an unrelated existing AppKit exception in `NSUpdateDynamicServices` before normal startup. The focused current-master runtime rerun used a test-only skip of that existing call. It is not part of this commit; the branch contains only the `udarwinfswatch.pas` change. The 2,640-transition stress run used the patch-only build without this test workaround.

## Commit

`b141512 FIX: avoid macOS FSEvents UI deadlock`

### DIFF
--- a/src/platform/unix/darwin/udarwinfswatch.pas
+++ b/src/platform/unix/darwin/udarwinfswatch.pas
@@ -133,7 +133,7 @@ private
   procedure handleEvents( const originalSession:TDarwinFSWatchEventSession );
   procedure doCallback( const watchPath:String; const internalEvent:TInternalEvent );
 
-  procedure updateStream;
+  procedure updateStream( const watchPaths:NSArray );
   procedure closeStream;
   procedure waitPath;
   procedure notifyPath;
@@ -615,16 +615,16 @@ begin
   event.Free;
 end;
 
-procedure TDarwinFSWatcher.updateStream;
+procedure TDarwinFSWatcher.updateStream( const watchPaths:NSArray );
 begin
-  if _watchPaths.isEqualToArray(_streamPaths) then exit;
+  if watchPaths.isEqualToArray(_streamPaths) then exit;
 
   closeStream;
 
   _streamPaths.release;
-  _streamPaths:= NSArray.alloc.initWithArray( _watchPaths );
+  _streamPaths:= NSArray.alloc.initWithArray( watchPaths );
 
-  if _watchPaths.count = 0 then
+  if watchPaths.count = 0 then
   begin
     _lastEventId:= FSEventStreamEventId(kFSEventStreamEventIdSinceNow);
     exit;
@@ -633,7 +633,7 @@ begin
   _stream:= FSEventStreamCreate( nil,
               @cdeclFSEventsCallback,
               @_streamContext,
-              CFArrayRef(_watchPaths),
+              CFArrayRef(watchPaths),
               _lastEventId,
               _latency/1000,
               CREATE_FLAGS );
@@ -657,6 +657,9 @@ end;
 procedure TDarwinFSWatcher.start;
 var
   pool: NSAutoReleasePool;
+  watchPaths: NSArray;
+  keepRunning: Boolean;
+  pathsChanged: Boolean;
 begin
   _running:= true;
   _runLoop:= CFRunLoopGetCurrent();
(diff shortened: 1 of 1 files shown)


rich2014:
@EricXh 

it's great!

there's something could be optimized. i'll merge it first, and then you can take a look.

would it be more appropriate to move the lock and loop from `start()` into `updateStream()`?

rich2014:
@alexx2000 

could you take a look at this?

1. Eric's patch largely avoids the mutual exclusion problem of long-running SMB I/O by reducing the lock range and using spin retries.

2. it is possible for the specific case of fswatch on SMB on macOS.

3. but a deeper and broader problem is that operations like changing paths running in the main thread, rather than a non-main thread.

4. it is the root cause of the whole DC UI being frozen by I/O.

5. is it possible to run these operations in a non-main thread? as the case of loading file list. that way, even if a certain I/O operation takes too long, it would only affect a single tab.

EricXh:
@rich2014 
Thanks! After some more testing, I should clarify the result:

The patch reliably removes the reproducible FSEvents lock contention in my stress tests. However, during normal navigation on my TrueNAS SMB share, I still occasionally see the same visible symptom: the whole UI freezes for about 5–10 seconds.

At this point I cannot say whether these remaining stalls still involve FSEvents or are caused by synchronous path-change operations such as stat, chdir, or the path availability check running in the main thread. So your broader concern seems very relevant.

Regarding the suggested cleanup: moving the snapshot, lock and retry loop into updateStream() sounds appropriate, provided that closeStream() / FSEventStreamInvalidate() continues to run outside the lock.

rich2014:
looking forward to your new patch :)

> Regarding the suggested cleanup: moving the snapshot, lock and retry loop into updateStream() sounds appropriate, provided that closeStream() / FSEventStreamInvalidate() continues to run outside the lock.

PR #3054, opened 2026-08-24
UPD: encapsulate macOS FSEvents stream updates

EricXh:
Follow-up to #3052, implementing the cleanup suggested in [this review comment](https://github.com/doublecmd/doublecmd/pull/3052#issuecomment-5391537523).

## Summary

- Move the watch-path snapshot, locking, and retry loop from `start()` into `updateStream()`.
- Have `updateStream()` return the watcher running state, leaving `start()` responsible only for the autorelease pool and run-loop lifecycle.
- Keep `closeStream()` / `FSEventStreamInvalidate()` outside `_lockObject`.

No functional change is intended. Each update still operates on an immutable snapshot, then rechecks the live paths under the lock and retries if they changed while stream teardown was in progress.

## Testing

- macOS arm64 Cocoa debug build
- macOS arm64 Cocoa release build
- 2,460 rapid directory transitions across several directories on a TrueNAS SMB share; no crash, hang, lost path update, or stuck UI
- Immediate shutdown after initiating a path change; the watcher thread exited and joined cleanly
- `git diff --check`

### DIFF
--- a/src/platform/unix/darwin/udarwinfswatch.pas
+++ b/src/platform/unix/darwin/udarwinfswatch.pas
@@ -133,7 +133,7 @@ private
   procedure handleEvents( const originalSession:TDarwinFSWatchEventSession );
   procedure doCallback( const watchPath:String; const internalEvent:TInternalEvent );
 
-  procedure updateStream( const watchPaths:NSArray );
+  function updateStream: Boolean;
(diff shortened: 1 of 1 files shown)


rich2014:
@EricXh 

merged, thx!

PR #3082, opened 2026-09-08
FIX: macOS drag&drop to external apps only delivering one file (multi-file selection)

vmurin:
## Problem

On macOS, dragging multiple selected files out of Double Commander into a modern Cocoa app (confirmed with WhatsApp for Mac) delivered only **one** file to the target; the same multi-selection dropped onto Finder/Desktop landed correctly, and no native "N items" drag badge appeared during the drag.

## Root cause

`TDragDropSourceCocoa.DoDragDrop` (`src/platform/udragdropcocoa.pas`) used the deprecated single-item drag call `-dragImage:at:offset:event:pasteboard:source:slideBack:`. It always creates exactly **one** drag item, with every selected path bundled into a single legacy `NSFilenamesPboardType` property-list array on that one item.

Finder still reads that whole legacy array, so drops onto Finder/Desktop worked. Modern AppKit apps instead enumerate the drag session's *per-item* pasteboard entries — with only one item, they only ever saw one file. The missing drag-count badge was the same symptom: AppKit only draws it for a real multi-item `NSDraggingSession`.

## Fix

Switch to the modern multi-item API: build one `NSDraggingItem` per file (each backed by its own file-URL pasteboard writer) and start the drag via `-beginDraggingSessionWithItems:event:source:`. This makes the drag genuinely asynchronous (the old call blocked until drop/cancel), so the drag-end event now fires from a new `NSDraggingSource` delegate's `draggingSession:endedAt:operation:` callback instead of synchronously right after starting the drag.

## Testing

Built and run on Apple Silicon (FPC 3.2.2 / Lazarus main, Cocoa widgetset). Manually verified:
- Multi-file drag from a Double Commander panel into WhatsApp for Mac — all files now attach, not just one.
- Same multi-selection dropped onto Finder/Desktop — still works, no regression.
- Native macOS drag-count badge and per-file icons now show during the drag, matching Finder's own behavior exactly.
- Single-file drag — no regression on either target.

### DIFF
--- a/src/platform/udragdropcocoa.pas
+++ b/src/platform/udragdropcocoa.pas
@@ -47,6 +47,46 @@ implementation
 uses
   CocoaAll, uDarwinUtil;
 
+type
+  { TCocoaDragSource }
+
+  { Bridges AppKit's NSDraggingSource callbacks back to a Pascal closure.
+    One instance is created per drag operation and released once the
+    drag session actually ends (draggingSession:endedAt:operation:). }
+  TCocoaDragSource = objcclass(NSObject, NSDraggingSourceProtocol)
+  public
+    DragEndEvent: uDragDropEx.TDragEndEvent;
+
+    function draggingSession_sourceOperationMaskForDraggingContext(
+      session: NSDraggingSession; context: NSDraggingContext): NSDragOperation;
+      message 'draggingSession:sourceOperationMaskForDraggingContext:';
+
+    procedure draggingSession_endedAtPoint_operation(
+      session: NSDraggingSession; screenPoint: NSPoint; operation: NSDragOperation);
+      message 'draggingSession:endedAtPoint:operation:';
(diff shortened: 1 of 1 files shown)


rich2014:
@vmurin 

thx for the patch!

there are a few points that need further consideration, but i prefer to merge it first, and then you can refine them:

1. because it was changed to asynchronous, an unnecessary `Reload()` was executed immediately in `TFileView.BeginDragExternal()`.

2. icons were added to all `DragItems`. if the number of `DragItems` is huge, this poses a certain risk. icons should only be necessary set for the first few `DragItems`.

3. converting `ScreenStartPoint` to `ViewPoint` seems unnecessary. setting a fixed `Origin` in `ItemFrame` should suffice. eg. `ItemFrame:= NSMakeRect(0, 0, 32, 32);`

Issue #3014, opened 2026-08-03
macOS Access violation


futurist:
### Version

| DC v1.1.32 gamma Rev. 830 -- aarch64-Darwin-cocoa
| macOS 26.5.2 | PID 3171

| DC v1.2.7 gamma Rev. 359 -- aarch64-Darwin-cocoa
| macOS 26.5.2 | PID 76493

OS INFO: MacBook Pro 16-inch, M5 Max

### Description

I've tried many version, constantly show the error:

```
--------------- 13-07-2026, 15:39:05 ---------------
| DC v1.1.32 gamma Rev. 830 -- aarch64-Darwin-cocoa
| macOS 26.5.2 | PID 3171
Unhandled exception: EAccessViolation: Access violation
  Stack trace:
  $000000010106E908 line 535, column 1 of uoperationsmanager.pas in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $00000001011C1568 line 231, column 6 of fFileOpDlg.pas in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100B26B70 line 2974, column 5 of include/control.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100D54F3C line 55, column 3 of include/buttoncontrol.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100D55844 line 169, column 3 of include/buttons.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100D58108 line 67, column 3 of include/bitbtn.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100D54DE0 line 21, column 5 of include/buttoncontrol.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $00000001009D756C in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100B25488 line 2304, column 24 of include/control.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100B165C8 line 5478, column 28 of include/wincontrol.inc in /Applications/Double Commander.app/Contents/MacOS/doublecmd
  $0000000100D7EEC4 line 114, column 36 of lclmessageglue.pas in /Applications/Double Commander.app/Contents/MacOS/doublecmd
[...]

PR #3083, opened 2026-09-09
FIX: macOS drag&drop - follow-up to #3082

vmurin:
Follow-up to #3082, addressing @rich2014's review comments.

1. `Reload` now runs when the drag ends, not right after starting it. Moved from `TFileView.BeginDragExternal` to `TFileViewWithMainCtrl.OnExDragEnd`. `DoDragDrop` only blocks on Windows and Qt — GTK (`gtk_drag_begin`) has always been asynchronous too, so this fixes GTK as well. The Windows transform-back branch keeps its order via `try/finally`.
2. Drag images only for the first 3 items, instead of one `NSWorkspace.iconForFile` per dragged file.
3. Fixed dragging frame. The old conversion was also wrong: `ScreenStartPoint` is LCL top-left origin, `convertRectFromScreen` expects Cocoa bottom-left.

Three more issues found on the way:

4. The session was started with whatever `currentEvent` happened to be — a key event when Command turns an internal drag into an external one.
5. `DoDragDrop` always returned False and never set `FLastStatus`.
6. The session delegate could call back into a file view freed while the drag was still running. Defensive, no hand reproduction.

Built and manually tested on Apple Silicon, Cocoa: multi-file drag into WhatsApp and onto Finder, Command during an internal drag, cancel, large selections.

Item 1 is in shared code, so it compiles everywhere, but only GTK and Cocoa change behaviour. On Windows and Qt the drag-end callback already runs synchronously inside `DoDragDrop`, so the reload still happens after the drop, just a few statements earlier. It was only built and run on macOS.

### DIFF
--- a/src/platform/udragdropcocoa.pas
+++ b/src/platform/udragdropcocoa.pas
@@ -224,7 +224,10 @@ begin
       DragItem.setDraggingFrame_contents(ItemFrame, ItemIcon);
     end
     else
-      DragItem.setDraggingFrame(ItemFrame);
+      // setDraggingFrame: keeps the item's default contents, which still get
+      // drawn. Clearing the contents explicitly is what leaves the trailing
+      // items out of the stack, the way Finder does it.
+      DragItem.setDraggingFrame_contents(ItemFrame, nil);
 
     DragItems.addObject(DragItem);
     DragItem.release;
--- a/src/platform/udragdropcocoa.pas
+++ b/src/platform/udragdropcocoa.pas
@@ -32,7 +32,11 @@ uses
 
 type
   TDragDropSourceCocoa = class(TDragDropSource)
-
+  private
+    { Called by the drag session delegate once AppKit reports the session
+      as finished. }
+    procedure DragSessionEnded(Succeeded: Boolean);
+  public
     function RegisterEvents(DragBeginEvent  : uDragDropEx.TDragBeginEvent;
                             RequestDataEvent: uDragDropEx.TRequestDataEvent;
                             DragEndEvent    : uDragDropEx.TDragEndEvent): Boolean; override;
@@ -64,7 +68,7 @@ type
     drag session actually ends (draggingSession:endedAt:operation:). }
   TCocoaDragSource = objcclass(NSObject, NSDraggingSourceProtocol)
   public
-    DragEndEvent: uDragDropEx.TDragEndEvent;
+    Owner: TDragDropSourceCocoa;
 
     function draggingSession_sourceOperationMaskForDraggingContext(
       session: NSDraggingSession; context: NSDraggingContext): NSDragOperation;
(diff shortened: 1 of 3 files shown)


src/platform/udragdropcocoa.pas
```diff
@@ -149,9 +162,13 @@ function TDragDropSourceCocoa.DoDragDrop(const FileNamesList: TStringList;
...
-    ItemIcon:= NSWorkspace.sharedWorkspace.iconForFile(StringToNSString(FileNamesList[I]));
-    ItemFrame:= NSMakeRect(ViewPoint.x - 16, ViewPoint.y - 16, 32, 32);
-    DragItem.setDraggingFrame_contents(ItemFrame, ItemIcon);
+    if I < MaxDragImageCount then
+    begin
+      ItemIcon:= NSWorkspace.sharedWorkspace.iconForFile(StringToNSString(FileNamesList[I]));
+      DragItem.setDraggingFrame_contents(ItemFrame, ItemIcon);
+    end
+    else
+      DragItem.setDraggingFrame(ItemFrame);
```
rich2014: `setDraggingFrame` and `setDraggingFrame_contents:` (passing `nil`) produce different results.

you can observe the difference, for example, by trying to drag 10 files in `Finder`.

rich2014 (changes requested):

src/platform/udragdropcocoa.pas
vmurin: Fixed in 8c74b4ac0 — now `setDraggingFrame:contents:` with `nil`.

src/platform/udragdropcocoa.pas
```diff
@@ -224,7 +224,10 @@ function TDragDropSourceCocoa.DoDragDrop(const FileNamesList: TStringList;
       DragItem.setDraggingFrame_contents(ItemFrame, ItemIcon);
     end
     else
-      DragItem.setDraggingFrame(ItemFrame);
+      // setDraggingFrame: keeps the item's default contents, which still get
```
rich2014: the code is correct, but the comments seem to be wrong. 

are you certain that the situation on your macOS matches what the comments describe?

which version of macOS are you using?

vmurin: Sorry — that was my assumption, never verified. Explanations removed, only what the code does is left. macOS 26.6.2, Apple Silicon.

rich2014:
@vmurin 

merging!

rich2014:
@vmurin 

could you test on macOS 27?

vmurin:
Not yet — I'm still on macOS 26.6.2 (Apple Silicon).

rich2014:
@vmurin 

could you test this issue: https://github.com/doublecmd/doublecmd/issues/3014?

PR #3087, opened 2026-09-11
ADD: honor GNOME ".hidden" files in the file list

p0n0marev:
## Summary
- Honor GNOME/Nautilus `.hidden` files when building the filesystem file list (Mantis [0002021](https://doublecmd.sourceforge.io/mantisbt/view.php?id=2021)).
- Names listed in a directory's `.hidden` file are treated as hidden/system files: they are filtered when *Show system/hidden files* is off, and dimmed when that option is on.
- Cached per directory and invalidated on reload.

## Test plan
- [ ] In a directory, create `.hidden` with a regular (non-dot) filename, e.g. `secret.txt`.
- [ ] With *Show system/hidden files* off, `secret.txt` should not appear in the panel.
- [ ] Toggle *Show system/hidden files* on: `secret.txt` should appear; with *Show hidden files dimmed* it should be dimmed.
- [ ] Edit `.hidden` and refresh the panel: the list should follow the new names.
- [ ] Files that start with `.` still behave as before.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/filesources/filesystem/ufilesystemfilesource.pas
+++ b/src/filesources/filesystem/ufilesystemfilesource.pas
@@ -88,6 +88,9 @@ type
     function GetRootDir: String; override; overload;
     function GetPathType(sPath : String): TPathType; override;
 
+    function IsSystemFile(aFile: TFile): Boolean; override;
+    function IsHiddenFile(aFile: TFile): Boolean; override;
+
     function CreateDirectory(const Path: String): Boolean; override;
     function FileSystemEntryExists(const Path: String; const Options: TFileSourceExistsOptions): TFileSourceExistsResult; override;
     function GetFreeSpace(Path: String; out FreeSize, TotalSize : Int64) : Boolean; override;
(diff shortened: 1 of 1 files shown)


src/filesources/filesystem/ufilesystemfilesource.pas
```diff
@@ -167,9 +170,89 @@ implementation
...
+    Index := DotHiddenCache.IndexOf(CachedPath);
+    if Index < 0 then
+    begin
+      if DotHiddenCache.Count >= DotHiddenCacheLimit then
+        ClearDotHiddenCache;
+      Names := LoadDotHiddenNames(CachedPath);
+      DotHiddenCache.AddObject(CachedPath, Names);
+    end
+    else
+      Names := TStringListEx(DotHiddenCache.Objects[Index]);
```
TheNicker, line 250: Does that mean that a cached `.hidden` file needs no processing if it's a cache hit ? 
Will DoReload be called if the `.hidden` file is modified while in cache ?

TheNicker:
This could be a nice addition even though it is not inherently a Linux feature but more of Linux desktop feature, so that might be subject to interpretation. 

AFAIK, `.hidden` exists beyond Gnome, basically any QT and GTK based environments, so the gravity of this PR is more than presented, might be worth updating the title and description.

PR #3090, opened 2026-09-13
FIX: Handle backslash-escaped quotes inside single-quoted command tem…

phhemberg:
…plates

When a filename contains an apostrophe (e.g. file's name.txt), the EscapeNoQuotes function escapes it as \'. However, when this escaped string is placed inside single quotes by the terminal command template (sh -c '{command}'), the parser treats \' as literal \ plus a closing quote, causing an EInvalidQuoting exception.

Fix: In RemoveQuotationOrSplitCmdLine, when inside single-quote mode and encountering \ followed by ', preserve both characters and do not close the quote. This allows the shell to correctly interpret the backslash escape while keeping the single-quoted context intact.

### DIFF
--- a/src/udcutils.pas
+++ b/src/udcutils.pas
@@ -786,7 +786,9 @@ begin
   sCommand := '';
   SetLength(Args, 0);
   QuoteChar := #0;
-  for I := 1 to Length(sCmdLine) do
+  I := 1;
+  while I <= Length(sCmdLine) do
+  begin
     case QuoteChar of
       '\':
         begin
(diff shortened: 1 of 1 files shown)


phhemberg:
Hallo all,
This is my first GIT commit ever - so if anything has been made the wrong way - pls. bare with me. The code/fix is legit and will fix the problem. 

The above code solves the quoting problem when running own scripts in toolbar with %t0 or &t1 parameter.

Thank's for at great peace of software.

PR #3094, opened 2026-09-14
Add: configurable drive selector positioning

TheNicker:
## Summary

Fixes two bugs and add a **Drives list position** setting under **Options → Layout**.

| Mode | Popup position |
| --- | --- |
| Legacy (default) | Below the requested panel's drive button. |
| File cursor (Active row) | At the focused file in the requested panel; falls back to the drive button when no visible file anchor is available. |
| Mouse pointer (Mouse cursor) | At the current mouse-pointer position. |

## Feature

- Remember the selected placement mode between sessions.
- In file-cursor mode, use the requested panel's active-file rectangle from the existing columns, brief or thumbnail view. Anchor the popup at its top edge, 20% across its visible width.

## Bug Fix

- Prevent clipping by the main window and correct screen-edge placement for selectors that fit within the monitor's work area.
- Emit one close notification on Escape or drive selection; previously these paths fired it twice.

## Notes

Previously, the visual hierarchy was `TfrmMain → TDrivesListPopup`: the grid was a child control of the main form, so any part outside the main form's client area was clipped. It is now `TDrivesListForm → TDrivesListPopup`, with the grid inside a separate borderless top-level window (`FHostForm`). This lets the selector extend beyond the main window's client area. Popups that fit within the monitor's work area are repositioned to keep them fully visible. The main form still owns the selector component for lifetime management.

- Keep the new host window sized to match the drive-list grid during live updates.
- Dismiss the selector before drive-selection callbacks, completed Main commands and chosen native menu commands so dialogs use the main window.
- Preserve the selector when a context menu is cancelled, and release mouse capture when it closes.

- Lists larger than the monitor's work area retain the existing clipping limitation.


Skif-off:
Please forgive my curiosity...

```pascal
array[TDrivesListPosition] of String =
    ('Legacy', 'ActiveRow', 'MouseCursor');
```
Why not 0, 1 & 2?

TheNicker:
Yeap I see, thanks for pointing this out.

The honest answer is intuition, by default I tend to use stable identifies for storage, unless there's a **real** reason not to. 

I can see that the entire XML configuration use automatic enum ordinals, so I'll update the code.

PR #3095, opened 2026-09-14
Fix: TSynUniSyn colors (Numbers/Punctuators/...) not persisted across restart

### LINKED ISSUE #2514: Not saving or applying highlighter editings
### Version

Double Commander
Version: 1.1.28 gamma
Revision: 787
Commit: b204fbb61
Build: 2025/08/10
Lazarus: 4.3.0.0
Free Pascal: 3.2.2
System: Mac OS X 26.0.0

### Description

Trying to change colors in highlighter on macOS, but DC not memorizing it.

DC installed by Homebrew.

xeningem:
## Summary

Fixes #2514 — custom colors for anything other than the `Strings` attribute in a UniLanguage-based highlighter (JSON, JS, Rust, Kotlin, Swift, C#, Go, PowerShell, QML, ...) silently reset to the shipped `.hgl` defaults on every restart, whether set through Options → Highlighters or by editing `colors.json` directly.

## Root cause

`TSynUniSyn.LanguageName` (`components/synunihighlighter/source/SynUniHighlighter.pas`) is a `class function` hardcoded to always return the literal string `"UniLanguage"`, the same value for *every* instance of the class regardless of which actual language (`Info.General.Name`) it represents.

`TdmHighl.CreateHighlighters` (`src/dmhigh.pas`) builds `SynHighlighterHashList` keyed by `HighLighter.LanguageName`:

```pascal
SynHighlighterHashList.Add(HighLighter.LanguageName, HighLighter);
```

Every `TSynUniSyn` instance (JSON, Rust, Kotlin, Swift, ...) therefore collides under the single key `"UniLanguage"`.

`TdmHighl.LoadUniColors` looks a highlighter up by its real name (e.g. `"JSON"`, the value stored in `colors.json` — the same name `SaveUniColors` already writes via `SynUniSyn.Info.General.Name`) through that same hash list:

```pascal
Highlighter:= TSynCustomHighlighter(SynHighlighterHashList.Data[AName]);
if Assigned(Highlighter) and (Highlighter is TSynUniSyn) then
  ...
```

Since no uni-highlighter is ever registered under its own name, this lookup always misses (`Highlighter = nil`) for every uni-highlighter, so `LoadUniColors` silently does nothing. The in-memory `TSynUniSyn` objects keep whatever colors were parsed straight from the `.hgl` file. On the next save, `SaveUniColors` (which iterates `SynHighlighterList` directly and does **not** need the hash list) faithfully re-serializes that untouched, still-default state — overwriting whatever the user had customized.

This explains the asymmetry originally reported in #2514: the top-level `Strings` range attribute happened to look like it persisted in some manual tests, while nested `Sets`/`KeyLists` (`Numbers`, `Punctuators`, `Keywords`) did not — in fact *none* of it was ever being loaded back; only the specific values a tester had most recently hand-edited into the file, right before checking, could appear "correct" without an intervening app restart having had the chance to overwrite them.

## Fix

Key `TSynUniSyn` instances in `SynHighlighterHashList` by their actual per-instance `Info.General.Name` instead of the always-constant `LanguageName`, matching what `SaveUniColors` already uses when writing `colors.json`.

## Test plan

- [x] Built Double Commander from source (macOS/cocoa) with the fix.
- [x] Isolated repro, no UI interaction involved:
  1. `./doublecmd --config-dir=/tmp/dc_test_cfg --no-splash`, quit immediately (SIGTERM) to generate a fresh default `colors.json`.
  2. Directly edit `colors.json`: set the `Dark` style's `JSON` uni-highlighter `Root.Sets[Name=Numbers].Attributes` to a custom color.
  3. Relaunch and quit again with zero UI interaction.
  4. Before the fix: the custom color reverted to the `.hgl` default. After the fix: the custom color survives the restart.
- [x] Also confirmed via targeted logging that, before the fix, `SynHighlighterHashList.Data['JSON']` (and every other uni-highlighter name) returned unassigned during `LoadUniColors`, and that after the fix the lookup succeeds.


alexx2000:
You are wrong, [LanguageName](https://github.com/doublecmd/doublecmd/blob/f3cbe6b4e09f2f581116f7f5e62656e8b91c26da/src/uhighlighters.pas#L368) is not a class function and returns a correct value for `TSynUniSyn`.

Lazarus 4.99 has `LanguageName` too, so it can be conflict. But anyway the fix will be different.

PR #3097, opened 2026-09-15
Fix: TSynUniSyn colors (Numbers/Punctuators/...) not persisted on LCL 4.99

### LINKED ISSUE #2514: Not saving or applying highlighter editings

xeningem:
## Summary

Supersedes #3095, which `@alexx2000` closed pointing out (correctly, on the LCL version he tested against) that the fix wasn't needed: `TSynCustomHighlighterHelper.LanguageName`, a DC-local class helper in `src/uhighlighters.pas`, already special-cases `TSynUniSyn` to return its real per-instance name.

That's exactly right on older/stable LCL. It stops being true on **LCL 4.99**, and #2514 is real there — just for a different reason than my original diagnosis.

## Root cause

Starting with LCL 4.99, the new base highlighter class DC's SynEdit fork now derives from, `TLazEditCustomHighlighter` (`lazarus/components/lazedit/lazedithighlighter.pas`), declares a **real** `LanguageName` property of its own:

```pascal
function GetInstanceLanguageName: string; virtual;
...
property LanguageName: string read GetInstanceLanguageName;
```

A genuinely declared class member always shadows a class helper method of the same name in Object Pascal, even with the helper's unit in scope. So on 4.99, every call to `HighLighter.LanguageName` in `src/dmhigh.pas` silently stops going through DC's `uhighlighters.pas` helper and falls through to `TLazEditCustomHighlighter`'s own default `GetInstanceLanguageName` → `GetLanguageName`, which for `TSynUniSyn` (`components/synunihighlighter/source/SynUniHighlighter.pas`) is hardcoded to the class-constant `"UniLanguage"` for every instance.

`TdmHighl.CreateHighlighters` keys `SynHighlighterHashList` by `HighLighter.LanguageName`, so on 4.99 every uni-highlighter instance (JSON, Rust, Kotlin, Swift, C#, Go, PowerShell, QML, ...) collides under that one `"UniLanguage"` key. `LoadUniColors`'s by-name lookup into that hash list then always misses and silently no-ops for every uni-highlighter, so any custom color on an attribute other than the top-level `Strings` range (`Numbers`, `Punctuators`, `Keywords`, `Remark`, ...) is never re-applied on load — the next save simply re-serializes the untouched `.hgl` defaults over the user's saved value.

Confirmed with an isolated repro (`--config-dir` pointing at a throwaway config, editing `colors.json` directly, launching and quitting with zero UI interaction, built against a local LCL 4.99 checkout): before this fix, a custom `Numbers`/`Punctuators` color for the JSON highlighter reverted to the `.hgl` default on every restart; after it, the custom value survives.

## Fix

Point fix, two files:

- **`components/synunihighlighter/source/SynUniHighlighter.pas`** — `TSynUniSyn` now overrides the *virtual instance method* `GetInstanceLanguageName` (guarded by `{$if lcl_fullversion >= 4990000}`, matching this file's existing use of that guard elsewhere) to return `Info.General.Name`, the same per-instance name `SaveUniColors`/`LoadUniColors` already key on. This makes `TLazEditCustomHighlighter`'s own `LanguageName` property correct per-instance on 4.99+ via normal virtual dispatch — no dependency on class-helper shadowing rules.
- **`src/dmhigh.pas`** — reverts the `TdmHighl.CreateHighlighters` special-case from #3095's commit back to a single `SynHighlighterHashList.Add(HighLighter.LanguageName, HighLighter)` call. It's now redundant everywhere: on 4.99 this PR's fix already makes `.LanguageName` correct at the source; on pre-4.99 LCL, `TSynCustomHighlighterHelper.LanguageName` was never shadowed, so the plain call already resolved correctly without it — exactly as `@alexx2000` said.

No other files are touched.

## Test plan

- [x] Built against a local LCL 4.99 checkout (`Lazarus: 4.99.0.0` in the About dialog / version banner).
- [x] Isolated repro, no UI interaction involved:
  1. `./doublecmd --config-dir=<tmp> --no-splash`, quit immediately to generate a fresh default `colors.json`.
  2. Directly edit `colors.json`: set the active style's `JSON` uni-highlighter `Numbers` (Set) and `Punctuators` (KeyList) `Attributes` to custom colors.
  3. Relaunch and quit again with zero UI interaction.
  4. Before this fix (i.e. on top of #3095's commit alone): reverts to `.hgl` defaults. After this fix: custom colors survive.
- [x] A/B-verified against the pre-#3095 baseline too (single unconditional `SynHighlighterHashList.Add(HighLighter.LanguageName, HighLighter)`, no special-case): on this same LCL 4.99 build, `HighLighter.LanguageName` logs as the literal `"UniLanguage"` for every `TSynUniSyn` instance, confirming the collision and the original diagnosis's mechanism — just requiring the LCL 4.99 condition to manifest.

Fixes #2514

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MvnqxCqnD6KppzAVbtYGuu


xeningem:
How it looks in the current snapshot on the dev
image

In the current branch
image

alexx2000:
I applied a more complete  fix.

PR #3100, opened 2026-09-16
FIX: file list - give up TFile ownership before the display list takes over

vmurin:
`MakeAllDisplayFileList()` hands each `TFile` to a `TDisplayFile`, which then owns it. `OwnsObjects` was cleared only after the whole transfer, so an exception midway left both lists owning the transferred files and the `finally` block freed them twice. Clearing it before the transfer turns that into a leak instead.

Found while looking into #3014. That crash is not reproducible here and this is not claimed to fix it — the double free is a separate defect, valid on its own.


rich2014:
@alexx2000 

Alex, could you look at this?

alexx2000:
> Clearing it before the transfer turns that into a leak instead.

So this fix replaces one bug by another (but less dangerous). I think this could be used as a temporary solution.

If original crash in the `MakeAllDisplayFileList` will be fixed then this leak will disappear.

vmurin:
Small clarification on "replaces one bug by another": the trade isn't an equal one. The double free was already there — this PR doesn't introduce a bug, it converts an existing one into a less dangerous form. On the normal path nothing changes at all; the difference only shows on the exception path, where heap corruption (freeing the same `TFile` twice, crashing anywhere later) becomes a few leaked objects in a process that is already unwinding.

That said, no reason to leave it at a trade. #3103 makes the handover exact — each `TFile` is cleared from the source list at the moment its `TDisplayFile` takes it, so there is neither a double free nor a leak, and the `OwnsObjects` flip goes away entirely. It's stacked on this branch, so it can either be merged after this one or taken instead of it.

PR #3101, opened 2026-09-16
FIX: macOS - drag and drop out of the panels is broken on macOS 27 (issue #2988)

### LINKED ISSUE #2988: critical issue with drag-and-drop functionality in Double Commander on macOS
### Version

Double Commander
Version: 1.1.32 gamma
Revision: 830
Commit: 187d1473a
Build: 2026/01/08
Lazarus: 4.5.0.0
Free Pascal: 3.2.2
System: macOS 27.0.0

### Description

Currently, I am unable to drag files from Double Commander and drop them anywhere outside the app.
Specifically:
•	Desktop/Finder: Dragging files to the Desktop or any Finder window does not work at all.
•	External Apps: Dragging files into web browsers (like Google Chrome) or professional software (like DaVinci Resolve) is completely ignored.

I'm on macos Golden Gate Beta 3

What I have tried so far:
	1.	Resetting system permissions via tccutil reset All com.company.doublecmd.
	2.	Giving Double Commander full "Accessibility" and "Full Disk Access" permissions in macOS System Settings.
	3.	Removing the quarantine flag using sudo xattr -rd com.apple.quarantine.
None of the above solved the issue. It seems like the app's drag-and-drop implementation (possibly related to how NSPasteboard / public.file-url is declared in the Lazarus/Pascal LCL wrapper) is failing to register the dropped files with the macOS window server or other Cocoa-based applications.
Are there any workarounds for this, or is this a known limitation of the current Cocoa/Qt5 interface on newer macOS versions?

vmurin:
On macOS 27 a drag out of a file panel completes the moment the cursor crosses the window boundary. Files land where the pointer left the window rather than where the button was released, and drops into another application's drop zone work only intermittently.

### What is wrong

The drag session never tracks the cursor. With the session instrumented, every attempt showed the same shape: it starts normally, receives exactly **one** `movedToPoint` callback, and ends — with the mouse button still physically down.

The cause is *when* the session is started. `DoDragDrop()` is reached only after LCL's internal drag has begun and the pointer has already left the window, through the internal-drag-then-transform dance in `TFileViewWithMainCtrl`. A session begun at that point is never handed the mouse. Started at the drag threshold instead, while the cursor is still over the panel, it tracks normally — 68 to 277 `movedToPoint` callbacks per drag where there had been one.

### What this changes

Starting the session early means it must also serve drops **inside** the application, since there is no longer an LCL drag to serve them. That is exactly what `CreateDragDropTarget()` could not do on Cocoa: it has no branch for that widgetset and returns the abstract dummy, so `OnExDragEnter`, `OnExDragOver`, `OnExDrop` and `OnExDragLeave` never fire, and incoming drops are served by a DARWIN-only workaround routing `TForm.OnDropFiles` through `TfrmMain.FormDropFiles`.

This adds the missing `TDragDropTargetCocoa`, wires the Cocoa branch into the factory so macOS takes the same path as Windows, GTK and Qt, and removes the workaround.

Drags originating in our own panels are recognised through `ExternalDragSourceControl` and keep the rules internal dragging has always had: `OnExDragOver` defers to `MainControlDragOver` rather than restating which file may be dropped onto which, and `OnExDrop` passes the source panel so the drop counts as internal.

### A second defect, found on the way

The session was being started on the wrong view. `TWinControl.Handle` is the scroll host wrapping the control, not the content view that is hit-tested and receives the mouse. Starting the session on the wrapper made Finder fail to read the dragged file — `invalidPathErr` in the CopyEngine log, surfaced to the user as *"an unexpected error occurred (error code -8060)"* — for any operation other than the move Finder picks by default.

This is why holding Option to copy a file out of Double Commander has been failing. It works now.

### Fixed as a consequence

Long-standing gaps on macOS rather than part of the regression, all closed by the target simply existing:

- a drop from Finder onto a subdirectory row lands in that subdirectory, not in the panel's current directory
- the target row is highlighted while dragging over it
- a drop onto an archive is handled

### On the drop target implementation

macOS requires the dragging-destination methods to live on the view's **class**. The panel's `NSView` is created by the LCL widgetset, which offers no per-control drop-target hook, so the methods are added to `NSView` through an Objective-C category and dispatched via a registry of registered views. Views that never called `registerForDraggedTypes` are not asked at all.

The solution is self-contained and requires nothing from other projects, but it does rely on a class we do not own: should LCL ever implement the same methods itself, the two definitions would conflict. This part therefore stays tied to how the widgetset develops and is worth re-checking when it is updated.

### Not addressed here

**Drop effect asymmetry.** Dragging out follows the receiving application's conventions (Finder: move by default, Option to copy); dropping in follows `gDefaultDropEffect` (copy by default, Shift to move). Pre-existing, and not something we can change without taking the system-standard modifiers away from the user.

### Testing

Verified by hand on macOS 27.0 (26A428), aarch64, Cocoa: drags out to Finder and to Chrome, panel-to-panel drags including onto subdirectories and archives, drops from Finder onto the panel, onto a subdirectory row and onto an archive, target highlighting, the rejection rules, and drops onto non-panel areas.

The Qt/macOS configuration was **not** built. All new code is behind `{$IFDEF LCLCOCOA}`, checked by reading, but that Lazarus install has only the Cocoa widgetset and neither CI nor the release scripts build Qt on macOS (`lcl=cocoa` throughout).

Closes #2988


rich2014:
@alexx2000 

i have reviewed the code, and it is ready to be merged.

there are a few areas that could be optimized later.

please take look, if you have no objections, i plan to merge it.

src/fileviews/ufileviewwithmainctrl.pas
```diff
@@ -1181,9 +1180,23 @@ procedure TFileViewWithMainCtrl.MainControlMouseMove(Sender: TObject; Shift: TSh
...
+            point without following the cursor. The session serves drops inside
+            the application too, so there is no separate internal drag here. }
+
+          // Restore selection of active file before the dragged list is built.
+          if (FSelectedCount > 0) and (not AFile.Selected) then MarkFile(AFile, True);
+
+          BeginDragExternal(AFile, FDragDropSource, FMainControlLastMouseButton,
+                            MainControl.ClientToScreen(Classes.Point(X, Y)));
+{$ELSE}
           MainControl.BeginDrag(False);
```
alexx2000, line 1196: I guess without internal dragging support the dropping onto toolbar (main/middle) will not work anymore.

vmurin: You're right. And it's not only the toolbar — dropping files on a toolbar button, on a tab header, or on a drive button is gone as well. All of them still work on Windows, GTK and Qt, so this would leave macOS as the odd one out, and it fails silently. I didn't test those cases, which is why I missed it.

One related change is intentional: a drop from Finder onto a non-panel part of the window no longer lands in the active panel. That's what the other widgetsets do, so I left it that way — say if you'd rather keep it.

Side note while I was there: on macOS a toolbar button made by dropping a document or a folder never worked anyway, only application bundles do. That's old and unrelated to this PR.

There seem to be three ways to close this. The toolbars, notebooks and drive bars can be registered as native drop targets, which restores all four, though it does widen the part of this PR that leans on an Objective-C category over `NSView`. The same can be done for tab headers and drive buttons only, leaving the toolbar out on macOS. Or all four can stay out, with a changelog note and a follow-up issue. Which of them is the right one, I can't decide.

